PR #2, opened 2023-04-08
Fixed spelling on exit message

GameTec-live:
Changed the exit message from "Bye, thanks for you.  ^.^ " to "Bye, thank you.  ^.^ ", which is a neater message.


xianglin1998:
image

PR #3, opened 2023-04-11
Added requirements.txt

GameTec-live:
Having a requirements.txt is never a bad idea, especially if the project isnt covered by the default libraries (eg: colorama, pyserial)
Its also recommended to keep this file up to date with all used libraries and having a recent ish version of the libraries used.

If theres something missing or the versions (somehow) arent fine, please, edit the file...


xianglin1998:
image

PR #7, opened 2023-04-30
Comment language

derGraph:
Changed the Comment language to English for better readability


GameTec-live:
maybe also translate the comments in the python files, would be nice... (i alrdy translated the .gitignore, so no need for that)

PR #12, opened 2023-05-31
[Fix] LF send edge issue

ShigemoriHakura:
when reenable lf timer, only half edge is sent, which causes the PM3 failes to get the correct start bit.


xianglin1998:
image

PR #13, opened 2023-06-01
Fixed Porxgrind typo

GameTec-live:
Fixed the USB descriptor typo from Porxgrind to Proxgrind

Im a typo developer ;)


Msprg:
Porxgrind 😂 why's it so funny to me...


Good catch~!

GameTec-live:
As stated on the discord server: one more typo and it will be fatal 🤣

PR #15, opened 2023-06-17
Expose nfc_tag_mf1_is_detection_enable and fix compilation on newer ARM compilator versions

Foxushka:


doegox:
Hello
```
CFLAGS += -Wno-array-bounds
CFLAGS += -Wno-sizeof-array-div
```
these don't fix the compilation they hide away potential problems that should be addressed.
Meanwhile, I suggest to rather use
```
CFLAGS += -Wno-error=array-bounds
CFLAGS += -Wno-error=sizeof-array-div
```
so warnings will still appear but without interrupting compilation.

Also, I'm not the maintainer, but I'd suggest to not propose different changes in one single PR, but separate proposals for CFLAGS, for typos and for stlink in separate PRs, so this is easier to discuss and accept them separately.

xianglin1998:
> 

I don't have sufficient experience in Github project management, which may require your help. I am also learning from you and the Iceman.

Foxushka:
Changed both to -Wno-error=

xianglin1998:
In order to reduce the confusion of commit records, I have manually merged them. Next time, your PR needs to be submitted according to Doebox's suggestion, which will make the git tree of the entire project clearer.

PR #27, opened 2023-07-18
fix LF EM emulation

doegox:
Some IDs are not emulated properly and are not detected properly e.g. with a Proxmark.
This patch removes the `m_is_send_reboardcast_last_edge` hack and the call to `ANT_NO_MOD` when checking the field presence.


xianglin1998:
Good!!!

PR #34, opened 2023-07-21
client: don't let recv thread crashing after hw dfu

doegox:
Add flag to cmd to tell to close connection nicely after having sent the cmd.
Useful e.g. for closing comm after sending `hw dfu`


xianglin1998:
Nice

PR #42, opened 2023-07-23
Upload DFU zip files to releases

Foxushka:
GitHub doesn't allow unauthorized users to download artifacts from actions. This PR will allow GitHub Actions to publish releases (pre-release) with .zip files for DFU (so users can flash it without GitHub account)

Required for https://github.com/GameTec-live/ChameleonUltraGUI/pull/52

### DIFF
--- a/.github/workflows/firmware.yml
+++ b/.github/workflows/firmware.yml
@@ -48,11 +48,6 @@ jobs:
     steps:
       - name: Check out the repo
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


doegox:
I merged this PR but it generates a "pre-release" for every commit, will it keep only the latest pre-release? else it will become soon hard to manage

gentilkiwi:
No full DFU ?

PR #45, opened 2023-07-28
Add readline import for Posix operating systems to handle arrow keys.

shallax:


iceman1001:
That improves it a lot :)

GarethWright:
Three lines of code to stop me screaming at my keyboard. Perfect!

PR #50, opened 2023-08-10
Add animation modes support

domints:
This change adds animation change support + ability to control it from CLI.
By default it uses full animation mode (as stock firmware) but it allows user to disable wake-up and sleep animations completely (mode 2) or use minimal animation (mode 1).

The reason for making this change is that waiting even for this single cycle of led swipe when triggered by reader makes it hard to use with some readers. Moreover, having such long animations is, to say the least, frustrating. 

Reason of having mode 1 is that when there is no animation at all when you wake Chameleon with button press, button up is registered as command to change slot, so slight delay helps alleviate that problem. In this mode there is no sleep swipe nor rf field wake-up animation. 

In both mode 1 and 2 there are animations for USB, slot change etc, I didn't change those as they didn't impact user experience.

This PR also adds 32 bit field to store user settings in flash - led animation uses 2 of them, next setting proposed on Discord is disabling bluetooth, which could also use one bit of this settings field.

### DIFF
--- a/firmware/application/src/settings.h
+++ b/firmware/application/src/settings.h
@@ -3,15 +3,17 @@
 
 #include <stdint.h>
 
+#include "utils.h"
+
 #define SETTINGS_ANIMATION_FULL 0
 #define SETTINGS_ANIMATION_MINIMAL 1
 #define SETTINGS_ANIMATION_NONE 2
 
-/*
- * bits [0-1]: animation config
- * bits [2-31]: reserved
- */
-typedef uint32_t settings_data_t;
+
+typedef struct ALIGN_U32 {
+    uint16_t version;
+    uint16_t animation_config : 2;
+} settings_data_t;
 
 void settings_load_config(void);
 uint8_t settings_save_config(void);
--- a/firmware/application/src/settings.h
+++ b/firmware/application/src/settings.h
@@ -1,0 +1,21 @@
+#ifndef SETTINGS_H
+#define SETTINGS_H
+
+#include <stdint.h>
+
+#define SETTINGS_ANIMATION_FULL 0
+#define SETTINGS_ANIMATION_MINIMAL 1
+#define SETTINGS_ANIMATION_NONE 2
+
+/*
+ * bits [0-1]: animation config
+ * bits [2-31]: reserved
+ */
+typedef uint32_t settings_data_t;
+
+void settings_load_config(void);
+uint8_t settings_save_config(void);
+uint8_t settings_get_animation_config();
+void settings_set_animation_config(uint8_t value);
+
+#endif
--- a/firmware/application/src/settings.h
+++ b/firmware/application/src/settings.h
@@ -5,19 +5,27 @@
 
 #include "utils.h"
 
-#define SETTINGS_ANIMATION_FULL 0
-#define SETTINGS_ANIMATION_MINIMAL 1
-#define SETTINGS_ANIMATION_NONE 2
+#define SETTINGS_CURRENT_VERSION 1
 
+typedef enum {
+    SettingsAnimationModeFull = 0,
+    SettingsAnimationModeMinimal = 1,
+    SettingsAnimationModeNone = 2,
+} settings_animation_mode_t;
 
 typedef struct ALIGN_U32 {
     uint16_t version;
-    uint16_t animation_config : 2;
+    uint8_t animation_config : 2;
+    uint16_t reserved0 : 14;
+    uint32_t reserved1;
+    uint32_t reserved2;
 } settings_data_t;
 
+void settings_init_config(void);
+void settings_migrate(void);
 void settings_load_config(void);
 uint8_t settings_save_config(void);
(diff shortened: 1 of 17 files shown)


GameTec-live:
yes please. Altough in my opinion it should start out as minimal or none  -\(°_°)/-

doegox:
maybe we can add some logic in mode 2 to avoid interpreting the wake-up button press as a slot change request.

GameTec-live:
> maybe we can add some logic in mode 2 to avoid interpreting the wake-up button press as a slot change request.

This could be argured both ways...
With no animations one could say that "waking from sleep" isnt "rewquired" and pressing the button should just immedietly perform the action. On the other hand that isnt consistent with the other modes...

firmware/application/src/settings.h
```diff
@@ -0,0 +1,21 @@
...
+
+#define SETTINGS_ANIMATION_FULL 0
+#define SETTINGS_ANIMATION_MINIMAL 1
+#define SETTINGS_ANIMATION_NONE 2
+
+/*
+ * bits [0-1]: animation config
+ * bits [2-31]: reserved
+ */
+typedef uint32_t settings_data_t;
```
augustozanellato, line 14: Probably a bitfield struct would be cleaner imho.
There's should also be a `version` field (8 bit should be enough? 16 to be sure?) so that settings can be cleanly upgraded.

domints:
Wow, nice, I didn’t know of bitfield strict. After the weekend I will
rework it this way :)
And version field - I can add it too, although I am not too sure if there’s
really need for it, as long as we ensure that default value will be 0x00
after firmware update it will take on this value.

W dniu sob., 12.08.2023 o 12:17 Augusto Zanellato ***@***.***>
napisał(a):

firmware/application/src/settings.h
domints: Wow, nice, I didn’t know of bitfield struct. After the weekend I will
rework it this way :)
And version field - I can add it too, although I am not too sure if there’s
really need for it, as long as we ensure that default value will be 0x00
after firmware update it will take on this value.

augustozanellato: > And version field - I can add it too, although I am not too sure if there’s really need for it, as long as we ensure that default value will be 0x00 after firmware update it will take on this value.

Version field is useful if a setting changes value range, meaning or gets removed

GameTec-live:
any ETA on when this might be merged?

firmware/application/src/settings.h
domints: Added version field and made it bitfield struct, thus resolving your comment :)

augustozanellato:
Here are some changed I'd suggest: https://pastebin.pl/view/c7147a50
Beware that it breaks compatibility with the previous pr version.
In short:
- a proper configuration handling mechanism has been implemented: a clean config will be generated if it can't be loaded from flash, migration support has been added, a command that resets config to default has been added
- the config struct size was properly aligned to 32 bits
- some space has been reserved for future settings
- an enum is now used for the animation mode

Personal observation: wouldn't it be more consistent if on inactivity shutdown there was a single pass of the animation on shutdown like there is for the wakeup caused by a button press?

doegox:
@augustozanellato globally your changes look nice IMHO and worth being added to this PR.
Beware of some weird stuff on the pad:
* lines 218-220 "SettingsAnimati"
* line 268 spurious chars `&#40;self, args: argparse.Namespace&#41;`

Add PACKED to the struct definition, else some compilers may use different layout:
```c
typedef struct ALIGN_U32 {
     uint16_t version;
    uint8_t animation_config : 2;
    uint16_t reserved0 : 14;
    uint32_t reserved1;
    uint32_t reserved2;
 } PACKED settings_data_t;
```

augustozanellato:
>     * lines 218-220 "SettingsAnimati"
> 
>     * line 268 spurious chars `&#40;self, args: argparse.Namespace&#41;`

I'm not sure what's happening there, but it seems like it's a bug in pastebin.pl

I upload a copy of the patchset that addresses your concern [here](https://gist.github.com/augustozanellato/45c38e50c42efbe5c910082a583a70eb)

doegox:
> I upload a copy of the patchset that addresses your concerns 

now merged in this PR, thanks

doegox:
I just added a fix for mode 2.
With mode 2 on a single button press, even long, it never takes it for a slot change request, so I'm not sure we need mode 1 at all...

domints:
So, is everything you guys wanted implemented or do you need anything changed from me? :D
Also, you're right, mode 1 seems unnecessary, but tbh I learned to like that single >half swipe on button wake-up :)
So I am voting to keep it :)

doegox:
ok let's merge, I can't live without it already :D

doegox:
@domints when I run `hw slot nick set -s 1 -st 1 -n test` I loose my animation settings and it goes back to mode 0.

YOBOAUSTIN (commented):
ok

PR #52, opened 2023-08-11
[Bugfix] Attempted fix of Offset issues and utilizing enums

F9Alejandro:
#### Created SlotNumber Enum with two staticmethods:
 - list: List all choices that can be made based on the name of the enum instead of value
 - fix: Takes the value and checks against an enum name that has been manipulated and converted to an int, if they match it outputs the value of the name/value enum pair

#### Added two staticmethods to TagSenseType Enum:
 - list: Same as what SlotNumber does above it lists all possible choices in a pretty print (not used for actual choice)
 - choices: Used for the actual choices and is only the value itself

#### All instances of slot_num/index have been set to the type enum of SlotNumber
 - Usage of SlotNumber.fix is used to convert from "name" to value of the enum pair
#### All instances of sensetag have been set to the type enum of TagSenseType
 - No further changes where made as it is still using value (no trickery here)

## Changes to chameleon_unit.py:
- Made changes to SlotIndexRequireUint to properly reflect changes made in cmd.py
- Made changes to SenseTypeRequireUint to properly reflect changes made in cmd.py

### Changes to be committed:
- modified:   software/script/chameleon_cli_unit.py
- modified:   software/script/chameleon_cmd.py


software/script/chameleon_cmd.py
```diff
@@ -42,6 +42,37 @@
...
+class SlotNumber(enum.IntEnum):
+    SLOT_1 = 0,
+    SLOT_2 = 1,
+    SLOT_3 = 2,
+    SLOT_4 = 3,
+    SLOT_5 = 4,
+    SLOT_6 = 5,
+    SLOT_7 = 6,
+    SLOT_8 = 7,
+    SLOT_NO = 8,
```
doegox, line 55: why do you need SLOT_NO ?

software/script/chameleon_cmd.py
```diff
@@ -267,7 +318,7 @@ def write_em_410x_to_t55xx(self, id_bytes: bytearray):
             data.extend(key)
         return self.device.send_cmd_sync(DATA_CMD_WRITE_EM410X_TO_T5577, 0x00, data)
 
-    def set_slot_activated(self, slot_index):
+    def set_slot_activated(self, slot_index: SlotNumber):
```
doegox, line 321: each such function is defined twice, make sure to fix the second one too (e.g. set_slot_activated is also at line 553).

software/script/chameleon_cmd.py
F9Alejandro: Take a look at the way how the choices for the CLI is selected. SLOT_NO ideally should never have to be used. This will also allow us to change the index without having to change what the end user sees. SLOT_NO could technically be any number.

software/script/chameleon_cmd.py
F9Alejandro: Those shouldn't need to be changed since they call but don't actually do data validation like the actual function itself.

software/script/chameleon_cmd.py
doegox: so why is it there ? everything runs perfectly fine without SLOT_NO, why adding it here ??
we're not in a C enum where you need a marker to know when you reached the end.

doegox:
I pushed some rewrites detailed in the commit comments.
@xianglin1998 you wanted to do some more tests before merge ?

xianglin1998:
I'm so busy, sorry. LoL.

PR #54, opened 2023-08-11
Add git version command

nemanjan00:
I think it would be useful to have hash of latest git commit

I do not own chameleon ultra yet, so I did not have a chance to test it yet


augustozanellato:
Tested, LGTM.

domints:
Due to #50 being just merged you need to get new command id for your command :)

nemanjan00:
Fixed command id and merged main back into my branch

PR #56, opened 2023-08-15
Implement get slot data

domints:
This PR is meant to add ability to read saved slot data.
At this point EM410x reading works.

I'm creating this draft PR to gather your opinion on whether fixing the mess in command IDs is an ok move (some HF commands were in LF space [5xxx]) and if changing `lf em sim --id [id]` to `lf em sim set --id [id]` and `lf em sim get` is ok move.

My next move will be adding support for reading out MF1 sectors from emulator memory.


domints:
I will modify this feature - using it the way apps use it (calling `hw slot change` and getting slot data) will kill the flash quickly - it has only 10000 to 20000 write cycles endurance :O

domints:
@doegox sorry, I was wrong, missed a line - it's storing this info only when command is issued or Chameleon goes to sleep. Either way, getting data by slot number seems more reasonable, so apps won't change user selected slot when they want to get slot data

doegox:
Ha yes as you prefer. It makes more sense to not have to change selected slot to read it indeed.

PR #61, opened 2023-08-17
Implement "hw slot info" to show which tag type loaded into all slots

Foxushka:


software/script/chameleon_cmd.py
```diff
@@ -19,13 +19,16 @@
...
 DATA_CMD_GET_DEVICE_ADDRESS = 1012
-DATA_CMD_GET_GIT_VERSION = 1017
 
 DATA_CMD_SAVE_SETTINGS = 1013
 DATA_CMD_RESET_SETTINGS = 1014
 DATA_CMD_SET_ANIMATION_MODE = 1015
 DATA_CMD_GET_ANIMATION_MODE = 1016
 
+DATA_CMD_GET_GIT_VERSION = 1017
+
```
augustozanellato, line 29: 1018 is missing

firmware/application/src/app_cmd.c
```diff
@@ -340,6 +340,23 @@ data_frame_tx_t* cmd_processor_slot_data_config_save(uint16_t cmd, uint16_t stat
     return data_frame_make(cmd, STATUS_DEVICE_SUCCESS, 0, NULL);
 }
 
+data_frame_tx_t* cmd_processor_get_activated_slot(uint16_t cmd, uint16_t status, uint16_t length, uint8_t *data) {
+    uint8_t slot = tag_emulation_get_slot();
+    return data_frame_make(cmd, STATUS_DEVICE_SUCCESS, 1, (uint8_t *)&slot);
```
augustozanellato, line 345: `(uint8_t *)` is useless as slot is an `uint8_t`

software/script/chameleon_cmd.py
Foxushka: 1018 used to get current selected slot

augustozanellato: > 1018 used to get current selected slot

Yeah, but it should probably be also implemented in the cli, just for the sake of completeness.

Foxushka: we have a lot of unimplemented methods in cli, method to get currently selected slot is useless if we can't even change it

doegox:
Nice!
Two suggestions:
* implement `-h`, every command should support -h even if it does not take other parameters, see e.g. `hw slot update -h`
* it would be nice to display nicknames with `hw slot info` when available.

Foxushka:
>  see e.g. `hw slot update -h`

Doesn't work, just executes command as usual

doegox:
?
```
[USB] chameleon --> hw slot update -h
 - Update config and data from device memory to flash success.
```

and I wish `hw slot info -h` returns also one line of explanation rather than the slot info dumps

Foxushka:
Check sources...
https://github.com/RfidResearchGroup/ChameleonUltra/blob/fc0e016c045b136bd8444811befe68b9d8f9722f/software/script/chameleon_cli_unit.py#L960

doegox:
Argh ok `hw slot update` was a bad example :D and I got tricked.
IMHO every command should support `-h` and `hw slot update` should be fixed too.
So a user who want to know about a command can always use `-h` without being afraid it will run immediately without knowing what it will do.

doegox:
But we can think of a more global fix later.
So I'll merge but if you could add nicknames later that would be great too!

PR #63, opened 2023-08-18
Add factory reset command

augustozanellato:
It's implemented as deleting all the FDS records present in flash, this way it won't need to be touched when new FDS records are introduced.
It's suggested to run it once because commit 997f58d0d0a6d12861eeae0150a3f7f03ba8eddb left some unused records around in the flash.


augustozanellato:
Note: `fds_delete_sync` was borked because file id and key weren't set so `fds_evt_handler` would never set the success flag, this pr should also fix that.

PR #64, opened 2023-08-19
Use single release for dev builds

augustozanellato:
Fix workflow using deprecated stuff.

For an example release see [here](https://github.com/augustozanellato/ChameleonUltra/releases/tag/dev)


Msprg:
Much more concise and releases now look much better as well! Good job :)

PR #66, opened 2023-08-19
CLI Overhaul

szymex73:
Since this is a PR with many changes (which wasn't the initial plan), I **highly encourage** testing it yourself and reading through all the changes.
There is a possibility that I missed something when merging upstream changes into this branch.

Of course, any reviews/comments are appreciated, some of the changes introduced for the purpose of increasing maintainability are based on personal opinions on how to structure something like this.

Note on compatibility:
In its current state, the PR does not work on versions of Python prior to 3.9 because of type hinting used, if this is a problem and updating Python is not desired, those can be removed to allow it to work on at least 3.8 (maybe 3.7, not tested).
It should also be noted that if this is merged, any currently open PRs that add new commands will have to be adapted.

### Replacing `cmd_positive` and `cmd_standard` with a unified `cmd`
This change removes the separate `PositiveChameleonCMD` class and unifies the error checking into `BaseChameleonCMD` via a new `expect_response` decorator.
The decorator behaves the same way as the previous `PositiveChameleonCMD.check_status` method and accepts either a single status or a list of status codes like the previous function call.

Example:
```py
# Previous
class BaseChameleonCMD:
    # [...]
    def scan_tag_14a(self):
        return self.device.send_cmd_sync(DATA_CMD_SCAN_14A_TAG, 0x00, None)

class PositiveChameleonCMD(BaseChameleonCMD):
    # [...]
    def scan_tag_14a(self):
        ret = super(PositiveChameleonCMD, self).scan_tag_14a()
        self.check_status(ret.status, chameleon_status.Device.HF_TAG_OK)
        return ret
```

```py
# New
class ChameleonCMD:
    # [...]
    @expect_response(chameleon_status.Device.HF_TAG_OK)
    def scan_tag_14a(self):
        return self.device.send_cmd_sync(DATA_CMD_SCAN_14A_TAG, 0x00, None)
```

This results in removing `BaseCLIUnit.cmd_positive` and `BaseCLIUnit.cmd_standard` properties in favor of a unified BaseCLIUnit.cmd`.

### Introduction of a `prompt-toolkit` based input system
This replaces the previous `input`/`readline` based prompt in favor of a `prompt-toolkit` based one that allows for autocomplete and more reliable history (with the possibility of having it saved/loaded from a file).

This also includes an autocomplete implementation for both commands and command arguments (thanks @s0ca for your implementation PoC, helped me make a less messy command completer).

### New way of registering commands
[...]

PR #68, opened 2023-08-20
Restore Posix shell history.

shallax:


doegox:
thanks but due to CLI rework by PR #66 this is not necessary anymore

PR #73, opened 2023-08-21
Add port auto detection on WSL1

### LINKED ISSUE #20: Port autodetect does not work on WSL1
https://github.com/RfidResearchGroup/ChameleonUltra/blob/939df3f0aa9525d9754a011228cc6c0bda087d8b/software/script/chameleon_cli_unit.py#L200

`Chameleon Connect fail: 'NoneType' object has no attribute 'startswith'`

Not working on `WSL1`, it seems @iceman1001 use a PowerShell hack to detect it under Windows env. then pass results to scripts

m-kozlowski:
fixes #20

### DIFF
--- a/software/script/chameleon_cli_unit.py
+++ b/software/script/chameleon_cli_unit.py
@@ -8,6 +8,7 @@ import timeit
 import sys
 import time
 import serial.tools.list_ports
+from platform import uname
 
 import chameleon_com
 import chameleon_cmd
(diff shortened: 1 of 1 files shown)


xianglin1998:
Has the solution been tested?

xianglin1998:
And you can record your modified at changelog.md, let's community know about your contribution.

Foxushka:
LOL how even it supposed to work, WSL is linux based, but you are calling powershell.exe

GameTec-live:
> LOL how even it supposed to work, WSL is linux based, but you are calling powershell.exe

microsoft tools are available in wsl 1/2, you could even call exporer.exe, etc, in wsl...

Foxushka:
image

m-kozlowski:
> LOL how even it supposed to work, WSL is linux based, but you are calling powershell.exe

the same way it works for PM3, Have you even tried it? o_O
BTW,  command is explorer.exe, not exprorer.exe.

doegox:
IMHO PR #73 and #74 should be merged.
First try powershell (that's the method we're using in pm3 since ever without much complain) and fallback on reckless active probing if powershell is not available and if user opt-in.

And you could try harder finding powershell.exe.
For info pm3 WSL1 method is sth like
```
    PSHEXE=$(command -v powershell.exe 2>/dev/null)
    # If it fails (such as if WSLENV is not set), try using the default installation path
    if [ -z "$PSHEXE" ]; then
        PSHEXE=/mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe
    fi

    # Finally test if PowerShell is working
    if ! "$PSHEXE" exit >/dev/null 2>&1; then
        echo >&2 "[!!] Cannot run powershell.exe, are you sure your WSL is authorized to run Windows processes? (cf WSL interop flag)"
        exit 1
    fi
    $PSHEXE -command "Get-CimInstance -ClassName Win32_serialport | Where-Object {\$_.PNPDeviceID -like '*VID_9AC4&PID_4B8F*' -or \$_.PNPDeviceID -like '*VID_2D2D&PID_504D*'} | Select -expandproperty DeviceID" 2>/dev/null
    DEV=$(echo $DEV | tr -dc '[:print:]')
    # note /dev/ttyS counterpart may take time to appear...
    DEV=$(echo $DEV | sed -nr 's#^COM([0-9]+)\b#/dev/ttyS\1#p')
```

Actually PM3 has one more method to support native Windows, with sth like
`wmic /locale:ms_409 path Win32_SerialPort Where "PNPDeviceID LIKE '%VID_9AC4&PID_4B8F%' Or PNPDeviceID LIKE '%VID_2D2D&PID_504D%'" Get DeviceID 2>/dev/null | awk -b '/^COM/{print $1}'`

For WSL2 I don´t know.

m-kozlowski:
> wmic /locale:ms_409 ...

under the hood it's the same as `Get-CimInstance`.
There's one slightly different and quite a bit faster, but I'm not sure if 100% reliable:
`Get-PnPDevice -Class Ports -PresentOnly | where {$_.Name -like "*(COM*)*"} | Select-Object FriendlyName,DeviceID`

Foxushka:
What if wundus installed not on C drive?

m-kozlowski:
If user has messed up PATH and non standard install location, then I'd say that's the corner case I don't care about. Just pass the port as argument

gentilkiwi:
Worked on my computer ;) 👍

PR #74, opened 2023-08-22
Poor man's WSL1 port prober

Foxushka:
1. Solution is shit, but microsoft deserves such ones
2. Works out of box (and well, at least works, yes, I'm about #73)
3. Might break other serial devices, but who cares? Disconnect them or don't use prober
4. Won't work on very old firmware


doegox:
please try revised #73 now merged, this should fix your issue.

doegox:
This PR is not mergeable as such anymore. Please test the latest code and open an issue or a PR if not satisfying.
Thank you

PR #76, opened 2023-08-22
Fix slot delete

Foxushka:


doegox:
can you show how the current main branch fails ?
when I try it seems to work fine:
```
[USB] chameleon --> hw slot list
 - Slot 1 data:
   HF: Empty - Mifare Classic 1k
   LF: hello - EM410X
[USB] chameleon --> hw slot delete -h
  -st number, --sense_type number
                        1 = LF, 2 = HF,
[USB] chameleon --> hw slot delete -s 1 -st 2
[USB] chameleon --> hw slot list
 - Slot 1 data:
   HF: Empty - Unknown
   LF: hello - EM410X
```

Foxushka:
Now try to delete LF dump, it won't allow you

doegox:
yes you can't delete the last sensetype, delete works only when both sensetypes are available.
```
[USB] chameleon --> hw slot list
 - Slot 1 data:
   HF: Empty - Mifare Classic 1k
   LF: hello - EM410X
[USB] chameleon --> hw slot delete -s 1 -st 1
[USB] chameleon --> hw slot list
 - Slot 1 data:
   HF: Empty - Mifare Classic 1k
   LF: hello - Unknown
```

doegox:
[USB] chameleon --> hw slot delete -h
Delete sense type data for a specific slot. **The slot needs to have the other sense type correctly configured, otherwise an error will be thrown.**

Foxushka:
I don't see any point of this restriction. How we can clean whole slot without factory reset?

doegox:
No idea, feel free to change it, but your fix doesn't seem adequate, it's not about HF/LF inversion.

doegox:
pinging @augustozanellato who made `delete`

PR #77, opened 2023-08-22
Fix setting slot nick

sgnusov:


doegox:
ha thanks! I failed at some merge conflict

Issue #78, opened 2023-08-22
Offline clone of LF UID


nemanjan00:
It would be nice to be able to use long-press to clone LF UID

komax74:
great feature

nemanjan00:
Closed by merging #79

PR #79, opened 2023-08-22
Offline clone of LF UID

nemanjan00:
This is fix for #78

### DIFF
--- a/firmware/application/src/app_main.c
+++ b/firmware/application/src/app_main.c
@@ -481,6 +481,37 @@ static void btn_fn_copy_ic_uid(void) {
 
     nfc_tag_14a_coll_res_entity_t* antres;
 
+    if(tag_type[1] == TAG_TYPE_EM410X) {
+        uint8_t status;
+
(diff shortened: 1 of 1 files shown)


firmware/application/src/app_main.c
```diff
@@ -481,6 +481,37 @@ static void btn_fn_copy_ic_uid(void) {
...
+        uint8_t id_buffer[5] = { 0x00 };
+        status = PcdScanEM410X(id_buffer);
+
+        if(status == LF_TAG_OK) {
+            tag_data_buffer_t* buffer = get_buffer_by_tag_type(TAG_TYPE_EM410X);
+            memcpy(buffer->buffer, id_buffer, LF_EM410X_TAG_ID_SIZE);
+            tag_emulation_load_by_buffer(TAG_TYPE_EM410X, false);
+
+            // keep reader mode or exit reader mode.
+            if (!is_reader_mode_now) {
```
doegox, line 505: you can move the test out of the if(status == LF_TAG_OK). Right now you're doing it in both branches of the if...

PR #80, opened 2023-08-22
CI refactor

augustozanellato:
Changes made:
 - CI properly supports being ran on PRs and shouldn't fail anymore on
   the RRG repo;
 - The nonsense of unzipping archives during release creation isn't
   needed anymore because now there's an artifact that holds all the
   files (both zip and others) that should be included in a release;
 - Softdevice and Application verification were re-enabled (as per
   @doegox suggestion);
 - Softdevice is now bundled in `-firmware` artifact;
 - `dev` release tag is now updated ad every push to main;
 - On PRs a comment linking to artifact is now added by CI after
   success, artifact links are done with nightly.link so that they also
   fork for unauthenticated users;

There are some quirks:
 - on_pr runs on pull_request_target because it needs write access to
   packages, but pull_request_target implies that GITHUB_SHA is set to
   the base repo HEAD commit, so the whole `checkout-sha` thing is
   needed. This also implies that if one of the workflow files are
   changed in a PR the changed versions won't be ran against that same
   PR.
 - Client building per OS isn't yet implemented but all the necessary
   scaffolding is there. At the moment the per-OS clients are just a zip
   file containing all the client source.

Examples:
  - PR comment: https://github.com/augustozanellato/ChameleonUltra/pull/14#issuecomment-1688737709
  - `dev` release: https://github.com/augustozanellato/ChameleonUltra/releases/tag/dev


doegox:
image

no need to cheat, nobody is putting pressure on you :D

doegox:
thanks !

Issue #81, opened 2023-08-22
Offline read status notification with LEDs


nemanjan00:
It would be cool if we could use LEDs to signal if offline clone worked

PR #82, opened 2023-08-22
Offline read status notification with LEDs

nemanjan00:
Fix for #81


doegox:
From a dual LF/HF slot, after searching (without tag), slot LED end up on blue instead of red.

nemanjan00:
I do not have blue hard-coded anywhere, so I have no idea what is happening

nemanjan00:
I can confirm the same thing is happening to me

doegox:
Blue led fixed in 7ed8005

PR #85, opened 2023-08-23
Fix stuff I broke yesterday

augustozanellato:
Fixes `Overlapping segments detected at address 1040384.ERROR: The hex files cannot be merged since there are conflicts.`
Was caused by `build.sh` overwriting `application.hex` with a version merged with bootloader settings, wasn't catched during testing because only clean builds were tested.
Fix involves using a temp directory for storing the binaries that will be put into `type-binaries.zip` so that the merged application can be renamed to `application.hex` without breaking successive builds.

`flash-dfu-*.sh` scripts have also been fixed to account for the new file names and to properly enter dfu even if script is ran from repo root (for example `firmware/flash-dfu-app.sh`)

### DIFF
--- a/firmware/flash-dfu-full.sh
+++ b/firmware/flash-dfu-full.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+if [[ $BASH_SOURCE = */* ]]; then
+  cd -- "${BASH_SOURCE%/*}/" || exit
+fi
+
 if ! ../resource/tools/enter_dfu.py; then
     echo "Wait for device to be off"
     echo "Press B and plug"
@@ -9,4 +13,19 @@ while :; do
   lsusb|grep -q 1915:521f && break
   sleep 1
 done
-nrfutil device program --firmware objects/dfu-full.zip --traits nordicDfu
+
+device_type=ultra
+lsusb | grep 1915:521f | grep -q ChameleonLite && device_type=lite
+
+echo "Flashing $device_type"
+
+dfu_package=objects/${device_type}-dfu-full.zip
+
+if [ ! -f $dfu_package ]; then
+    echo "DFU package for $device_type not found, aborting."
+    echo "Build firmware using CURRENT_DEVICE_TYPE=$device_type firmware/build.sh"
+    exit 1
+fi
+
+nrfutil device program --firmware $dfu_package --traits nordicDfu
+
--- a/firmware/build.sh
+++ b/firmware/build.sh
@@ -60,15 +60,19 @@ set -xe
     --merge \
     settings.hex \
     application.hex \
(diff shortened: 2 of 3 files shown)


nemanjan00:
I am not a huge fan of having multiple places to put what device you have. 

Perhaps we should do Makefile with build and flash targets that sources Makefile.defs and passes device type as argument to flash?

augustozanellato:
> I am not a huge fan of having multiple places to put what device you have.
> 
> Perhaps we should do Makefile with build and flash targets that sources Makefile.defs and passes device type as argument to flash?

You only need to define it once while you're compiling. Flash scripts automatically detect the connected device type and pick the correct dfu package

firmware/flash-dfu-full.sh
```diff
@@ -9,4 +13,19 @@ while :; do
   lsusb|grep -q 1915:521f && break
   sleep 1
 done
-nrfutil device program --firmware objects/dfu-full.zip --traits nordicDfu
+
+device_type=ultra
```
nemanjan00, line 17: You have device type hardcoded here

augustozanellato: Check the line below :)

augustozanellato: defaults to ultra, if a lite is found it sets device_type to lite

nemanjan00:
Even the build.sh is not really using Makefile.defs, and it has its own way to set default

firmware/flash-dfu-full.sh
nemanjan00: Oh, sorry, my mistake

augustozanellato:
The makefile pulls the device type from an environment variable, build.sh just does some more things like packaging dfu etc.
Default is to build everything for ultra, if you want to build for lite you'd only need to run `CURRENT_DEVICE_TYPE=lite firmware/build.sh` without editing anything.

nemanjan00:
Maybe that is something that should be documented.

I missed ?=, thanks for clarification

doegox:
I'm in favor of replacing build.sh by a general Makefile, and load user prefs from a file not part of the repo (but template prefs file available in the repo).
So only one place to define prefs, and once you've set your prefs for your device, you don't have to worry anymore. And no risk to commit prefs by accident.
Just didn't have time for that yet.

nemanjan00:
My confusion started from assuming prefs are in single file like with proxmark3

doegox:
yes I'd like to make it the same as for pm3

doegox:
still WIP or ready ?

augustozanellato:
> still WIP or ready ?

Forgot to rename the pr, now it's ready.

Issue #87, opened 2023-08-23
Long-press key binding


nemanjan00:
It would be very useful to have long-press key binding so we can have more offline functions

PR #88, opened 2023-08-23
Add long press

nemanjan00:
Solutin for #87

### DIFF
--- a/firmware/application/src/app_main.c
+++ b/firmware/application/src/app_main.c
@@ -1,6 +1,7 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/types.h>
 
 #include "nordic_common.h"
 #include "nrf.h"
@@ -41,8 +42,16 @@ NRF_LOG_MODULE_REGISTER();
 
 // Defining soft timers
 APP_TIMER_DEF(m_button_check_timer); // Timer for button debounce
+
+static uint32_t m_last_btn_press = 0;
+
+static bool m_is_btn_long_press = false;
+
 static bool m_is_b_btn_press = false;
 static bool m_is_a_btn_press = false;
+
+static bool m_is_b_btn_release = false;
+static bool m_is_a_btn_release = false;
 
 // cpu reset reason
 static uint32_t m_reset_source;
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -51,5 +51,6 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
  - Fixed compilation errors with GCC 12 (@Foxushka)
  - Added documentation for JLink (@xianglin1998)
  - Added support for ST-Link and debugging documentation (@derGraph)
+ - Added support for long-press of buttons
```
doegox, line 54: add your gh handle and move the entry on the top, not bottom :) thanks!

software/script/chameleon_cli_unit.py
```diff
@@ -1330,6 +1333,7 @@ class HWButtonSettingsSet(DeviceRequiredUnit):
 
     def args_parser(self) -> ArgumentParserNoExit:
         parser = ArgumentParserNoExit()
+        parser.add_argument('-l', '--long', type=int, required=True, help="1 is Long or 0 Short", choices=[0, 1])
```
doegox, line 1336: I'd rather see --long as a flag without arg and optional. So short by default unless `--long`.
what do you think?

xianglin1998: me too.

CHANGELOG.md
nemanjan00: Done

software/script/chameleon_cli_unit.py
nemanjan00: I agree. Done

firmware/application/src/app_main.c
```diff
@@ -152,12 +161,41 @@ static void timer_button_event_handle(void *arg) {
...
+        }
+    }
+
+    if (nrf_gpio_pin_read(pin) == 0) {
+        uint32_t now = app_timer_cnt_get();
+        uint32_t ticks = app_timer_cnt_diff_compute(now, m_last_btn_press);
+
+        uint32_t time = ticks * ((APP_TIMER_CONFIG_RTC_FREQUENCY + 1 ) * 1000 ) / APP_TIMER_CLOCK_FREQ;
+
+        bool is_long_press = time > 1000;
```
doegox, line 182: sorry one more comment :)
I was about to suggest to create a macro as this can be useful elsewhere, but there is one for the inverse so you could replace by
```suggestion
        bool is_long_press = time > APP_TIMER_TICKS(1000);
```

nemanjan00:
@doegox, I think there is something wrong with CI. It marked the wrong commit as failed

PR #94, opened 2023-08-25
Fix compilation problem error: a label can only be part of a stateme…

gentilkiwi:
Fix compilation problem `error: a label can only be part of a stateme…nt and a declaration is not a statement`

Please avoid creating variable in middle of functions, and in this particular case double `uint8_t status;`.

Can be valid for some compiler, but does not help to read the code.


doegox:
Note for later: the error occurs on ARM GCC v8, and v10, but it's fine under v12

PR #96, opened 2023-08-26
Start working on TCP comms

szymex73:
This change will allow for connecting to a TCP server that exposes the chameleon serial interface, similar to how the PM3 CLI allows for connecting to a TCP server exposing a Proxmark. This should allow for the use of the CLI on devices like Android phones with Termux.

Currently, it's working for communicating with a device but needs a bug fixed with a thread locking up when trying to close the socket.

### DIFF
--- a/software/script/chameleon_cli_main.py
+++ b/software/script/chameleon_cli_main.py
@@ -73,7 +73,7 @@ class ChameleonCLI:
 
         :return: current cmd prompt
         """
-        device_string = f"{colorama.Fore.GREEN}USB" if self.device_com.isOpen(
+        device_string = f"{colorama.Fore.GREEN}USB" if self.device_com.is_open(
         ) else f"{colorama.Fore.RED}Offline"
         status = f"[{device_string}{colorama.Style.RESET_ALL}] chameleon --> "
         return status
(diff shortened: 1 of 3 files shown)


doegox:
Just sharing some thoughts, feel free to ignore, I don't want to make your life harder neither unless you think it's worth investigating...

To also support BLE in CLI, the way seems to use Bleak, which is based on asyncio. If we move to a unified asyncio model for transports, there is also pyserial-asyncio.
Now for TCP, if you use Streams it's also asyncio.

So, maybe should we investigate moving to a unified asyncio API for transport?
(I must confess I've always more difficulties thinking with async than sync, but well...)

GameTec-live:
> Just sharing some thoughts, feel free to ignore, I don't want to make your life harder neither unless you think it's worth investigating...
> 
> To also support BLE in CLI, the way seems to use Bleak, which is based on asyncio. If we move to a unified asyncio model for transports, there is also pyserial-asyncio. Now for TCP, if you use Streams it's also asyncio.
> 
> So, maybe should we investigate moving to a unified asyncio API for transport? (I must confess I've always more difficulties thinking with async than sync, but well...)

migrating to async might be worth it to prevent future issues, etc.

szymex73:
It would require some rearchitecturing but I do agree that going async might be a better choice with the transaction model we currently have. Could take a look at that pretty soon but can't say for sure

GameTec-live:

PR #101, opened 2023-08-26
Fix a bug in multiline command implementation

szymex73:
This fixes a bug where after a KeyboardInterrupt the application would crash since it would try to split a string that does not exist.


doegox:
Thanks! for my info, how did you trigger the bug? when I did try ctrl-d or ctrl-c, it was quitting without crash

szymex73:
Both Ctrl+C and Ctrl+D would trigger it on my Linux host

Issue #41, opened 2023-07-23
Some suggestions for firmware compatibility.


xianglin1998:
1. The part related to the communication protocol cannot be changed at will.
2. Those involving communication command codes and communication return parameters cannot be changed at will.
3. When adding any command, its function definition should be considered, and the functions related to HF should not be placed in the Device category.
4. If you need to delete the command, you can only mark it for deletion, remember not to rearrange the communication command code.
5. In any case, the client must be able to obtain the software version code for compatibility processing.
6. The current firmware version lacks many commands, which will lead to the client's functions being too simple or too difficult to use. It is necessary to wait for the v1.1 version to be released when the firmware is relatively complete, and update iteratively and perform compatibility processing from the v1.1 version.
7. After the v1.1 version of the firmware is released in the future, changes that affect compatibility should be more strictly reviewed to prevent the new firmware from being unable to run on the old client.

Note: We are currently in the stage of ChameleonUltra's firmware function construction and community ecological construction, and we can temporarily ignore the strict restrictions on adding/modifying commands.

xianglin1998:
Repositories related to Chameleon Ultra:

[ChameleonUltraGUI (Flutter)](https://github.com/GameTec-live/ChameleonUltraGUI) By @GameTec-live

doegox:
Can you elaborate what you mean by "the functions related to HF should not be placed in the Device category." and where it is in the code ?

doegox:
"new firmware from being unable to run on the old client" => do you want to be always capable to run all newer fw with any old client ? That means potentially being stuck with some initial choices that cannot be changed anymore.
On the proxmark3, we ask ppl to run fw and corresponding client version, and the client is able to detect automatically version mismatches and alert the user on the need to reflash the firmware, for example.
I can think of running a fw newer than the client only for things like an outdated GUI or mobile application not properly maintained.

xianglin1998:
> Can you elaborate what you mean by "the functions related to HF should not be placed in the Device category." and where it is in the code ?

The CMD code define at [Here](https://github.com/RfidResearchGroup/ChameleonUltra/blob/main/firmware/application/src/data_cmd.h) (Just to categorize the functions of CMD.)

xianglin1998:
[...]

PR #103, opened 2023-08-27
Change Chameleon Lite PID to 0x8787

Foxushka:
This will give us ability to separate ultra/lite without ugly hacks like `if "ChameleonUltra" in productName` (and as productName not even available in Web Serial API)
udev rule updated too, lite should appear in /dev/lite-X

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Changed Chameleon Lite PID to `0x8787` (@Foxushka)
  - Security BLE implemented (@xianglin1998)
  - Added `hw settings blekey` to get and set ble connect key (@xianglin1998)
  - Added `hw ble bonds clear` to delete all ble bonds (@xianglin1998)
--- a/firmware/application/src/sdk_config.h
+++ b/firmware/application/src/sdk_config.h
@@ -6420,7 +6420,7 @@
 // <i> Selected Product ID
 
 #ifndef APP_USBD_PID
-#define APP_USBD_PID 0x8686
+#define APP_USBD_PID DEVICE_PID
 #endif
 
 // <o> APP_USBD_DEVICE_VER_MAJOR - Major device version  <0-99>
--- a/firmware/common/device_info.h
+++ b/firmware/common/device_info.h
@@ -5,9 +5,11 @@
 #if defined(PROJECT_CHAMELEON_ULTRA)
 #define DEVICE_NAME_STR         "ChameleonUltra"
 #define DEVICE_NAME_STR_SHORT   "CU"
+#define DEVICE_PID              0x8686
 #elif defined(PROJECT_CHAMELEON_LITE)
 #define DEVICE_NAME_STR         "ChameleonLite"
 #define DEVICE_NAME_STR_SHORT   "CL"
+#define DEVICE_PID              0x8787
 #else
 #error "Unknown device name?"
 #endif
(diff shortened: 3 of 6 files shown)


doegox:
Rather than creating different PIDs every time a new device variant will emerge, and having to look for each of them, the plan was more to let the compatible devices reporting their own capabilities on the fly (capabilities which depend on the hardware but also possibly on firmware modules being enabled or not), and the client to adapt, as done in the proxmark ecosystem. See https://github.com/RfidResearchGroup/ChameleonUltra/issues/41#issuecomment-1646852672.
Still to be implemented, I agree.

Foxushka:
From this side it doesn't help either. As I said Web Serial API doesn't expose product name, so there is no way to know what device connected (Ultra or Lite)

doegox:
today easiest way is just to send a request to enter reader mode. if status NOT_IMPLEMENTED it's a Lite.
I'm changing now slightly the logic in the mode command, so Lite accepts a "mode set -e", just because there is no reason to refuse such command, even if it does nothing...

Foxushka:
Ultra and Lite have different product name, different BLE name, why they should have same PID?
We workaround detection by running write to t5577, but this still need connection to device to execute. Better change at start of development so it won't break anything

doegox:
Product names, BLE names, are just names. Today 2 variants, tomorrow 10 variants, with third parties rolling out theyr somehow compatible hardware. We don't want to have to track all their PIDs.

You will always have to connect the first time anyway, right?
Ok right now the workaround requires entering reader mode, which has no effect on Lite but has effect on Reader so that's not ideal. Before implementing a full list of capabilities we can just report the device model e.g. in a extra data of GET_APP_VERSION or in new GET_HW_VERSION so at first connect, you can query the device and know what it is.
Would it be good enough?

Foxushka:
> You will always have to connect the first time anyway, right?

No, there is no need to connect to device in main menu where we show chameleon images based on it's type. We don't probe/connect to chameleons on that moment. After connection separating ultra and lite is easy

doegox:
I'm lost, are you talking about discovery over BLE or USB?
If about BLE how USB PID can help?

Foxushka:
Over USB on web browser through Web Serial API that doesn't expose product name

doegox:
so in the other PR you wanted to send blindy bytes to any available serial device just to see if it was a chameleon and now, when you see a device with the right VID/PID, you don't want to send a single command to it to query its capabilities?

doegox:
BTW over Web Serial API, at least for what I know on Brave, your client cannot get a list of devices.
The user will see the list of his serial device *including names* and will select which one he connects to your client.

Foxushka:
> so in the other PR you wanted to send blindy bytes to any available serial device just to see if it was a chameleon and now

No, what you mean with other PR? We can separate is it chameleon or not chameleon device. But we can't separate ultra and lite before connecting to it and querying something. And when you have both lite and ultra chameleons connected... Good luck finding correct one

doegox:
image

doegox:
> > so in the other PR you wanted to send blindy bytes to any available serial device just to see if it was a chameleon and now
> 
> No, what you mean with other PR? 

your "slow probe_wsl"

Foxushka:
> The user will see the list of his serial device including names and will select which one he connects to your client.

Yes, same in chrome. But you can pair multiple devices.
image

And after connecting most than you can see is 
image

And if you think :1 will be ultra, you are wrong, it is lite

Foxushka:
> slow_probe_wsl

slow_probe_wsl is WSL1 fix, it has no relation to this PR or any issues I fixing with this PR

Foxushka:
> And if you think :1 will be ultra, you are wrong, it is lite

Or no, false positive? It is actually lite... Okay

doegox:
sorry but you ask for big changes that won't scale in the future just for a little GUI annoyance.
if I have 2 ultra connected, I will have the same problem anyway.

Foxushka:
> big changes

I don't think PID change at early stage of development is big change

doegox:
it is because it means every time a device with a new form factor will emerge you will expect yet another new PID for it, so you can display its pic.
Look, we managed to live with Proxmark3 for 15 years with *many* hardware variants and always the same VID:PID. Nobody complains.

Foxushka:
If there was proxmark web client that actually display image of your proxmark3... It would be a cool but useless feature.

Well, at least I tried, now we have a place to send users who will complain about that

PR #119, opened 2023-09-01
Update firmware.md

mitmarcus:
explanation on write modes provided by GameTec_live

### DIFF
--- a/docs/firmware.md
+++ b/docs/firmware.md
@@ -103,10 +103,10 @@ The Chameleon also shows the following LED effects:
 The device enters sleep mode after about 5s unless it is plugged in USB or if a client is connected over BLE. You can use the buttons to wake it up again. You can also press quickly a button during the sleep animation to keep the device awake.
 
 ## Write Modes
-- Normal: Behaves like any normal card
-- Decline: Read only card
-- Decive: Accept write commands but dont change any data (reader thinks write was sucesful but when reading nothing changed)
-- Shadow: Accept writes but revert when device goes to sleep (reader can read and write like a normal card but it reverts when the chameleon goes to sleep) 
+- **Normal**: Behaves like any normal card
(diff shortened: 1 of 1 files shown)


docs/firmware.md
```diff
@@ -102,6 +102,11 @@ The Chameleon also shows the following LED effects:
 
 The device enters sleep mode after about 5s unless it is plugged in USB or if a client is connected over BLE. You can use the buttons to wake it up again. You can also press quickly a button during the sleep animation to keep the device awake.
 
+## Write Modes
+- Normal: Behaves like any norml card
```
GameTec-live, line 106: norml? not normal?

mitmarcus: normal, normal

docs/firmware.md
```diff
@@ -102,6 +102,11 @@ The Chameleon also shows the following LED effects:
 
 The device enters sleep mode after about 5s unless it is plugged in USB or if a client is connected over BLE. You can use the buttons to wake it up again. You can also press quickly a button during the sleep animation to keep the device awake.
 
+## Write Modes
+- Normal: Behaves like any normal card
+- Decline: Read only card
+- Decive: Accept write commands but dont change any data (reader thinks write was sucesful but when reading nothing changed)
+- Shadow: Accept writes but revert when device goes to sleep (reader can read and write like a normal card but it reverts when the chameleon goes to sleep)
```
doegox, line 109:
```suggestion
## Write Modes
- **Normal**: Behaves like any normal card
- **Denied**: Read-only card, send NACK to write attempts
- **Deceive**: Accepts write commands but don't change any data (reader thinks write was successful but when reading back, nothing changed)
- **Shadow**: Accepts writes but reverts changes when device goes to sleep (reader can read and write like a normal card but changes are kept in RAM and are lost when the chameleon goes to sleep) 
```

PR #132, opened 2023-09-07
block_count can't be more than 0xFF, simplify code

Foxushka:


doegox:
Can't we have 256 blocks to be written on a 4k DW ?

Foxushka:
> Can't we have 256 blocks to be written on a 4k DW ?

Good point, but we can't get all blocks in one message anyway. I've tested it now, we might get only 32 blocks per message. Trying to get more than 32 blocks will lead to transfer error. Should we add 32 block limit?

doegox:
ok it's clearer now, thanks!

PR #140, opened 2023-09-09
Solved bug that makes hf mf eread command crashes

aenigma-es:
When use the "hf mf eread" command via CLI it crashes as bytearray() do not likes the params


doegox:
thank you!

PR #149, opened 2023-09-12
Update nomenclature technical_whitepaper.md

aramova:
Simulation and emulation used within same document while having no distinctive technical differentiating usage, propose changing all instances to Emulate/Emulation


doegox:
thanks

PR #147, opened 2023-09-11
Next release v2.0.0. Clarify protocol. Disruptive changes: see below

doegox:
This huge commit tries to enhance several things related to the fw/cli protocol.
Generally, the idea is to be verbose, explicit and reuse conventions, in order to enhance code maintainability and understandability for the other contributors.

docs/protocol.md got heavily updated

Many commands have been renamed for consistency. you are invited to adapt your client for easier maintenance

Guidelines, also written in docs/protocol.md "New data payloads: guidelines for developers":
- Now protocol data exchanged over USB or BLE are defined in netdata.h as packed structs and values are stored in Network byte order (=Big Endian)
- Command-specific payloads are defined in their respective cmd_processor handler in app_cmd.c and chameleon_cmd.py
- Define C `struct` for cmd/resp data greater than a single byte, use and abuse of `struct.pack`/`struct.unpack` in Python. So one can understand the payload format at a simple glimpse.
- If single byte of data to return, still use a 1-byte `data`, not `status`.
- Use unambiguous types such as `uint16_t`, not `int` or `enum`. Cast explicitly `int` and `enum` to `uint_t` of proper size
- Use Network byte order for 16b and 32b integers
  - Macros `U16NTOHS`, `U32NTOHL` must be used on reception of a command payload.
  - Macros `U16HTONS`, `U32HTONL` must be used on creation of a response payload.
  - In Python, use the modifier `!` with all `struct.pack`/`struct.unpack`
- Concentrate payload parsing in the handlers, avoid further parsing in their callers. This is true for the firmware and the client.
- In cmd_processor handlers: don't reuse input `length`/`data` parameters for creating the response content
- Avoid hardcoding offsets, use `sizeof()`, `offsetof(struct, field)` in C and `struct.calcsize()` in Python
- Use the exact same command and fields names in firmware and in client, use function names matching the command names for their handlers unless there is a very good reason not to do so. This helps grepping around. Names must start with a letter, not a number, because some languages require it (e.g. `14a_scan` not possible in Python)
- Respect commands order in `m_data_cmd_map`, `data_cmd.h` and `chameleon_cmd.py` definitions
- Even if a command is not yet implemented in firmware or in client but a command number is allocated, add it to `data_cmd.h` and `chameleon_cmd.py` with some `FIXME: to be implemented` comment
- Validate data before using it, both when receiving command data in the firmware and when receiving response data in the client.
- Validate response status in client.

Disruptive changes:
[...]

PR #151, opened 2023-09-13
Proper 14443-4a support

augustozanellato:
Automatically send RATS to 14443-4a tags
Add ATS copy to long press quick clone
Fix long press quick clone supporting only 4b uid

### DIFF
--- a/software/script/chameleon_cstruct.py
+++ b/software/script/chameleon_cstruct.py
@@ -20,9 +20,11 @@ def parse_14a_scan_tag_result(data: bytearray):
     """
     return {
         'uid_size': data[10],
-        'uid_hex': data[0:data[10]].hex(),
+        'uid_hex': data[0:data[10]].hex(' '),
         'sak_hex': hex(data[12]).lstrip('0x').rjust(2, '0'),
-        'atqa_hex': data[13:15].hex().upper()
+        'atqa_hex': data[13:15].hex(' '),
+        'ats_size': data[270],
+        'ats_hex': data[15:15 + data[270]].hex(' '),
     }
 
 
--- a/firmware/application/src/app_cmd.c
+++ b/firmware/application/src/app_cmd.c
@@ -491,13 +491,17 @@ data_frame_tx_t *cmd_processor_get_mf1_anti_coll_data(uint16_t cmd, uint16_t sta
     if (tag_type[0] == TAG_TYPE_UNKNOWN) {
         return data_frame_make(cmd, STATUS_PAR_ERR, 0, data); // no data in slot, don't send garbage
     }
-    uint8_t responseData[16] = {};
+    picc_14a_tag_t taginfo;
     nfc_tag_14a_coll_res_reference_t *info = get_saved_mifare_coll_res();
-    memcpy(responseData, info->uid, *info->size);
-    responseData[10] = *info->size; // size is 2 byte len, but...
-    responseData[12] = info->sak[0];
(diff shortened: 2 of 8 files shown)


software/script/chameleon_cstruct.py
```diff
@@ -20,9 +20,11 @@ def parse_14a_scan_tag_result(data: bytearray):
     """
     return {
         'uid_size': data[10],
-        'uid_hex': data[0:data[10]].hex(),
+        'uid_hex': data[0:data[10]].hex(' '),
         'sak_hex': hex(data[12]).lstrip('0x').rjust(2, '0'),
-        'atqa_hex': data[13:15].hex().upper()
```
augustozanellato, line 25: removed `.upper()` because it's called anyways by the caller, added space separator for clarity

software/script/chameleon_cstruct.py
```diff
@@ -20,9 +20,11 @@ def parse_14a_scan_tag_result(data: bytearray):
     """
     return {
         'uid_size': data[10],
-        'uid_hex': data[0:data[10]].hex(),
+        'uid_hex': data[0:data[10]].hex(' '),
```
augustozanellato, line 23: added space separator for clarity

doegox:
:grimacing: sorry.
If you're not in a hurry we can merge it manually after the massive PR, so this can have its own commit.
Good news is that there will be a new common format for tag info for HF14A_SCAN, MF1_GET_ANTI_COLL_DATA, MF1_SET_ANTI_COLL_DATA and ATS is already foreseen in it. 
cf https://github.com/RfidResearchGroup/ChameleonUltra/blob/wip_protocol_refresh/docs/protocol.md#2000-hf14a_scan

augustozanellato:
My fault, should have checked the open PRs before opening one :)

> If you're not in a hurry we can merge it manually after the massive PR

Definitely not in a hurry, feel free to take your time with the protocol refactoring. In the meantime I drafted this PR.

doegox:
I pushed your changes as https://github.com/RfidResearchGroup/ChameleonUltra/pull/147/commits/09eb0d19f450fba9c9f05aac6fdf3ed23c5b14e6 on the other PR.
Changes:
- on 7-byte UID, RATS was sent before cascade lvl2
- don't return errors on ATS, just log them, as we still want to get some info about the tag. The client can do the sanity checks.

augustozanellato:
> I pushed your changes as [09eb0d1](https://github.com/RfidResearchGroup/ChameleonUltra/commit/09eb0d19f450fba9c9f05aac6fdf3ed23c5b14e6) on the other PR. 
Ok, I’ll close this one then. 

> Changes:
> 
> * on 7-byte UID, RATS was sent before cascade lvl2

Not sure if there were issues while importing my commit onto your branch, but on the PR RATS was correctly sent after cascade, see [here](https://github.com/RfidResearchGroup/ChameleonUltra/blob/68d465d867945b88cc420f728223f8ff6f3f66f3/firmware/application/src/rfid/reader/hf/rc522.c#L583), RATS is sent outside the cascade loop.

> * don't return errors on ATS, just log them, as we still want to get some info about the tag. The client can do the sanity checks.

Good idea, I’d document that clients are expected to validate ATS.

doegox:
> Not sure if there were issues while importing my commit onto your branch, but on the PR RATS was correctly sent after cascade

ha yes, sorry, I copy/pasted too early above "}"

PR #170, opened 2023-10-12
Fix factory reset hang/wdt reset

augustozanellato:
fds_util event handler was ignoring all the events related to peer_manager files and that was causing a deadlock while waiting for record deletion that in turn caused a wdt reset, usually this resulted in a partially done factory reset.
The fix adds a `ignore_pm` flag to op_info struct that gets cleared by fds_wipe allowing for a true reset to happen.


doegox:
ha fantastic! I got the deadlock occasionally without managing to pinpoint the cause.

PR #174, opened 2023-10-17
Fix ATS handling in tags that NAK RATS

augustozanellato:
`pcd_14a_reader_ats_request` didn't check for NAK so in case the tag NAKd the RATS it would return a zero length ATS which would cause an underflow in `pcd_14a_reader_scan_once`, that in turn resulted in `HF_ERR_ATS` being returned due to an invalid ATS length.


doegox:
thanks !

PR #178, opened 2023-10-21
Change git version to version tag instead of dev tag.

taichunmin:
Change git version from `dev-5-g74f4eb8` to `v2.0.0-101-g74f4eb8`


doegox:
thanks!

PR #180, opened 2023-10-25
Added support for mifare classic value block operations

taichunmin:
Added support for mifare classic value block operations

### DIFF
--- a/software/script/chameleon_cmd.py
+++ b/software/script/chameleon_cmd.py
@@ -4,9 +4,9 @@ from typing import Union
 
 import chameleon_com
 from chameleon_utils import expect_response
-from chameleon_enum import Command, Status, SlotNumber, TagSenseType, TagSpecificType
-from chameleon_enum import MifareClassicDarksideStatus
-from chameleon_enum import ButtonType, ButtonPressFunction
+from chameleon_enum import Command, SlotNumber, Status, TagSenseType, TagSpecificType
+from chameleon_enum import ButtonPressFunction, ButtonType, MifareClassicDarksideStatus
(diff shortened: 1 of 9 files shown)


taichunmin:
I didn't know how to implement send bytes to tag and then wait for response nack or timeout (which means success) in mf value operation.

截圖 2023-10-25 上午9 08 00

software/script/chameleon_cmd.py
```diff
@@ -277,6 +277,29 @@ class CStruct(ctypes.BigEndianStructure):
...
+        :param operator:
+        :param operand:
+        :param dst_block:
+        :param dst_type:
+        :param dst_key:
+        :return:
+        """
+        data = struct.pack('!BB6sBiBB6s', src_type, src_block, src_key, operator, operand, dst_type, dst_block, dst_key)
+        resp = self.device.send_cmd_sync(Command.MF1_MANIPULATE_VALUE_BLOCK, data)
+        resp.data = resp.status == Status.HF_TAG_OK
```
doegox, line 300: beware now it's resp.parsed = ...

docker-compose.yml
doegox, line 1: unrelated, please use another PR if you want to add it and explain why
please remove from this PR

taichunmin: OK, I will remove the file before merge

software/script/chameleon_cmd.py
taichunmin: Fixed

doegox:
I didn't test specifically the new value block operations but I did some tests to make sure existing 14a support for raw, mfu etc was still working fine, so good enough for a merge ;)

PR #199, opened 2024-01-18
Added command to check keys of multiple sectors at once

taichunmin:
- CLI Usage:

- CLI Preview (Normal):

- CLI Preview (Tag removed):


xianglin1998:
nice

PR #205, opened 2024-02-25
cli: Fix cmd_parameter generation for nested for python 3.9

PetePriority:
I noticed that when running the CLI with python 3.9, the `type_target` enum would be converted to a string representation of the enum, instead to its numerical value, i.e., `MfcKeyType.A` instead of `96`. This caused the execution of the nested binary to fail.


PetePriority:
It still works on python 3.11.8. 
But this is expected: the nested binaries expect an integer as their argument. So `int(type_target)` makes sure the enum is converted to an int before the implicit conversion to a string in the format string `f"...{int(type_target)}..."`

xianglin1998:
Good!!!

PR #228, opened 2024-08-17
Add functionality behind --dic argument in CU CLI

LupusE:
As someone wanted to use the --dict in CLI and got an error, the source shows the functions `load_key_file` and `load_dict_file` are not defined.
After some tests, I believe the functionality does not need to be an own function, it is easier to read and understand in one block while parsing the keys.

The `if ... elif ... else` construct is not very elegant, but works better for the different types of comparing as a `match ... case`. It is good to extend, if a .dic file (or single keys from a file) needs to be converted.

### DIFF
--- a/software/script/chameleon_cli_unit.py
+++ b/software/script/chameleon_cli_unit.py
@@ -52,14 +52,6 @@ def check_tools():
         print(f'{CR}Warning, tools {", ".join(missing_tools)} not found. '
               f'Corresponding commands will not work as intended.{C0}')
 
-dict_keys = set()
-def load_dic_file(filename, keys):
-    #print("Read directory from", filename) ## debug
-    for dict_key in filename.readlines():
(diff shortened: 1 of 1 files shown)


doegox:
hello, you need to merge your last 2 commits and force-push on the MR branch, else the venv files will still land into the repo history. 
Something like
* `git reset HEAD^`
* `git commit -a --amend`
* `git push --force`

LupusE:
Thanks for the hint. I've tried ... 
Maybe I need to cancel the PR and restart. The issue came up because Github flagged my account a few month ago. At this time some repos gone wild. 
I even wonder why this change was transmitted to the RRG repo. I wanted to upload this only to my fork.

I hope it is clean, now. Else I will do a full reset, tomorrow.

doegox:
Yes the venv are now completely gone, thanks.
Now someone has to find time to fix the CI issues first :sweat_smile: , nothing to do with your changes.

PR #233, opened 2024-09-15
Update build_client.yml

Fearless-Squirrel:


GameTec-live:
What is your goal with this? why are you trying to get the runners IP address?

PR #239, opened 2024-11-19
`hf mf elog --decrypt` skip records with found keys

taichunmin:


GameTec-live:
why do you always do force pushes? lol

taichunmin:
@GameTec-live because I try to prevent three-way merge.

LupusE:
I would prefer a clean three-way-merge, because of the changes in software/script/chameleon_cli_unit.py.
But I have no prove that it necessary. If you say it won't affect main, it is fine for me.

GameTec-live:
one disadvantage of force pushes and more or less doing a one commit pr tho is that you lose the entire history and make bisection harder

PR #262, opened 2025-06-26
Fix incorrect slot LED indication when switching via CLI

### LINKED ISSUE #160: Problem with led on slot change
There is some problem with current slot indication while using in the same time the usb cli and the buttons:

On firmware version v2.0.0 do:

0. go to using buttons
1.  first slot is red
2. press button to next slot - slot 2 is green
3. press button to next slot - slot 3 is blue

While changing slots using buttons the colors are always in my case the same (1: red, 2: green: 3 blue). Now lets connect to cli:

4. using buttons change current slot to 3 (led 3 is blue)
5. `hw slot change -s 1` - slot has changed to 1 but now slot 1 led is red and slot 3 is red (should be slot 1:red, 3:off)
6. `hw slot change -s 2` - slot has changed to 2 but now slot 2 led is red and slot 3 is red (should be slot 2:green, 3 off)
7. `hw slot change -s 3` - slot has changed to 3 but now slot 3 led is red (not blue as it should to be)
8. Press button to previous slot, the slot 3 icon changes to blue and immediately it changes to slot 2 with green light (it looks like the led colors are corrected to the valid state)
9. `hw slot change -s 1` - slot has changed to 1 but now slot 1 led is red and slot 2 is red (should be slot 1:red, 2 off)

unkernet:
Previously, using `hw slot change -s N` via the CLI would cause inconsistent LED colors:
- The LED for the newly selected slot turned on, but old slot LEDs remained lit.
- LED colors did not reflect the actual slot configuration (e.g., slot 3 LED showed red instead of blue).

Changes:
- Used `get_color_by_slot()` to determine correct color per slot based on enabled tag types.
- Modified `get_color_by_slot()` to also check if LF/HF sensing is enabled.

Switching slots via CLI now behaves consistently with button-based slot changes.

Closes #160

### DIFF
--- a/firmware/application/src/app_cmd.c
+++ b/firmware/application/src/app_cmd.c
@@ -24,11 +24,11 @@
 NRF_LOG_MODULE_REGISTER();
 
 
-static void change_slot_auto(uint8_t slot) {
+static void change_slot_auto(uint8_t slot_new) {
+    uint8_t slot_now = tag_emulation_get_slot();
     device_mode_t mode = get_device_mode();
-    tag_emulation_change_slot(slot, mode != DEVICE_MODE_READER);
-    light_up_by_slot();
-    set_slot_light_color(RGB_RED);
+    tag_emulation_change_slot(slot_new, mode != DEVICE_MODE_READER);
+    apply_slot_change(slot_now, slot_new);
 }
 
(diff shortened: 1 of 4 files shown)


LupusE:
The change looks good to me. I tested it at my CU and the Fix worked. I also was able to reproduce the issue, before upgrading.

GameTec-live (approved):
LGTM and seems to work

PR #268, opened 2025-08-02
Automatic run make_style.sh on PR 

Foxushka:
As this project not mostly maintained, its worth running style on each PR, as this rarely done by anyone

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -1,0 +1,21 @@
+name: Code style
+
+on:
+  pull_request: null
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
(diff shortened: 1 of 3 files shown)


GameTec-live:
hm, what about using ruff for the python part?
Add it to the recommended extensions and run ruff format on pr?

LupusE:
In general I understand the urge of a unique style. I saw a lot of tabs and removed spaces in the last few PR. There are changes for a better readability, but makes the focus on the change harder while reviewing.

Do I understand correct, that this change will just cleanup/unify the code, don't touch functionality?

In my little experience, this kind of automatic is working at a somewhat clean/readable code and the CU code is far away from it.
But I would give it a shot.

Foxushka:
> Do I understand correct, that this change will just cleanup/unify the code, don't touch functionality?

Yes, it should not change anything

> There are changes for a better readability, but makes the focus on the change harder while reviewing.

In first PR where formatting is introduced yes, then all PRs will be formatted and there won't be 1000 lines changes, just what PR author edited

GameTec-live:
That mostly has been resolved now. For the c/c++ side were still searching for good solutions. Id be interested in clang-format as that seems to be industry standard...

PR #267, opened 2025-08-02
feat: add lf HIDProx read, t55xx write, emulate function

TeCHiScy:
code quaility:
- consistance: simulation -> emulation, label -> tag
- machine translated unreadable comments are changed to native human readable english
- convert legacy `nrf_drv_xxx` macro usage to new `nrfx_xxx` macro usage
- make `byte_mirror` a common util, and share its definitions between hf & lf

fix:
- fixed a issue that cause em410x write args not verified in python cli

logic:
- newly added cli command includes: `lf hid prox read`, `lf hid prox write`, `lf hid prox econfig`. `hw slot list` is also updated to show HIDProx type tag
- state machine in em410x machester demodulator is simplified/made clear, extends to support 16/32/64 bitrate
- various wiegand formats of HIDProx are supported
- goertzel algorithm is used in  HIDProx FSK2a demod (bitrate rf/50)
- lf read is refactored using stream/feed pattern to boost scan speed
- t55xx write is refactored to share same logic between em410x & HIDProx
- lf emulating is refactored to use PWM peripheral, allowing more card type to be added

other:
- this PR enables a common pattern for others to add another FSK/Manchester modulated lf tags

closes: #212 
closes: #210

PR #263, opened 2025-07-06
Key recovery via backdoor for static encrypted nonce cards

Foxushka:

PR #275, opened 2025-08-10
Restore Python 3.9 compatibility for Chameleon CLI

unkernet:
After the last few changes (#263, #267), the minimum Python version required to run the Chameleon CLI effectively became **3.11**, due to the use of:
* The Python 3.10+ type hint syntax `int | None`
* `int.from_bytes()` calls without the required `byteorder` argument (mandatory in < 3.11)

However, the CLI documentation still states Python 3.9 as the minimum supported version.

This PR restores compatibility with Python 3.9 by:
* Replacing `int | None` with `Union[int, None]` for type hints.
* Adding an explicit `byteorder='big'` argument to all `int.from_bytes()` calls.

This way, the documented Python 3.9 requirement remains accurate.
Alternatively, if Python 3.11+ is intended to be the new baseline, the documentation should be updated accordingly — but I suggest keeping compatibility where possible.

### DIFF
--- a/software/script/chameleon_cli_unit.py
+++ b/software/script/chameleon_cli_unit.py
@@ -421,7 +421,7 @@ class LFHIDIdArgsUnit(DeviceRequiredUnit):
         return parser
 
     @staticmethod
-    def check_limits(format: int, fc: int | None, cn: int | None, il: int | None, oem: int | None):
+    def check_limits(format: int, fc: Union[int, None], cn: Union[int, None], il: Union[int, None], oem: Union[int, None]):
         limits = {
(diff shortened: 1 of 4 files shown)


LupusE (approved):
Based on the upcoming EOL of Python 3.9 (10-2025), I suggest to step up to Python 3.10. Reference: https://www.python.org/downloads/
I have no point why 3.13 should not be possible, but at this status of the project, maybe a too huge step. I am open for suggestions.

I approve this PR, because I don't see any huge impact in readability or speed. Although I would also update the documentation to minimum 'Python 3.10'.

PR #281, opened 2025-08-11
fix error response of cmd MF1_CHECK_KEYS_ON_BLOCK

taichunmin:
When I implement the MF1_CHECK_KEYS_ON_BLOCK command in my JavaScript SDK, the error response for the command is incorrect.

/cc @Foxushka


Foxushka (approved):

LupusE (approved):
It is already approved by fuxoshka, so as a formality I approve as well :)
Looks valid to me!

PR #282, opened 2025-08-19
Powershell build script

misch2:
Add a Windows Powershell build script.

To run this:
```powershell
.\build.ps1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Added Powershell build script
  - Added UV, formatter and linter. Contribution guidelines.
  - Extend max packet data size from 512 to 4096 bytes (@Foxushka)
  - HID Prox support (@TeCHiScy)
--- a/firmware/build.ps1
+++ b/firmware/build.ps1
@@ -1,0 +1,85 @@
+# build.ps1
+
+$ErrorActionPreference = "Stop"
+
+# Change to script directory
+Set-Location -Path $PSScriptRoot
+
+$softdevice = "s140"
+$softdevice_version = "7.2.0"
+$softdevice_id = "0x0100"
+
+# TODO: find a way to manage this automatically, I don't want to rely on action build #.
+$application_version = 1
+$bootloader_version = 1
+
+$device_type = $env:CURRENT_DEVICE_TYPE
+if (-not $device_type) { $device_type = "ultra" }
+
+switch ($device_type) {
+  "ultra" { $hw_version = 0 }
+  "lite"  { $hw_version = 1 }
+  default {
+    Write-Error "Unknown CURRENT_DEVICE_TYPE $device_type, aborting."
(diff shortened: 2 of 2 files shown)


LupusE:
I don't know if we need a PowerShell build script. Does it simplify anything?

Line 30: I would add a -confirm:$false
Line 33: You executing `make -j` without testing if it exists. (same for nfrutil ... maybe a link to the wiki, if not found)
Line 73: `(Split-Path $tmp_dir)` is redundant to `| Split-Path` from line 72, isn't it?
Line 77,79,83: Same as Line30

misch2:
Hi LupusE, thank you for the review!

As for a need for this script: It makes it possible to build the firmware on Windows as AFAIK there's no easy way to get the bash running there. But if I have missed something and the PowerShell script is not needed I can close the pull request.

Lines 30 etc.: fixed

Line 33 etc.: The bash.sh doesn't check for an existence of these commands either so I didn't considered it necessary. The `$ErrorActionPreference = "Stop"` at the start of the script makes it exit with error if any of those commands are not found like in this example:
```
> .\build.ps1
Building firmware for ultra (hw_version=0)
Invoke-Expression : The term 'make' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or 
if a path was included, verify that the path is correct and try again.
At build.ps1:33 char:1
```

LupusE:
I am not a windows user. But as far as I understand you need to make sure the build environment is set up. So the bash is just one more dependency?

The script exits, yes. But what now? If there is already a helper script, it would be helpful to help the user:

```powershell
try {
    Get-Command -Type Application nrfutil
} catch [System.Management.Automation.CommandNotFoundException] {
    Write-Output "Please install nrfutils, see https://github.com/RfidResearchGroup/ChameleonUltra/wiki/development#install-nrf-tools for further instructions." 
} catch {
    Write-Output "I also don't know how this situation could happen. Please open an issue at https://github.com/RfidResearchGroup/ChameleonUltra/issues"
}
```

I have no windows near me, so I don't know if this is actually working. Just to illustrate the idea.
You could also write a try/catch around the whole nrfutil block ...

misch2:
Hi, yes, this script works, I'm using it to build the firmware on Windows 11.

I walked through the instructions on https://github.com/RfidResearchGroup/ChameleonUltra/wiki/development#Prerequisites-for-compiling to install the necessary software.
Everything there is working on Windows up to the "Compiling the code" part. There it suggests running `build.sh` (which is not possible on Windows) or _"try to execute its steps manually if your platform is not yet properly supported."_. 

My intention was to help with this final step, i.e using `build.ps1` to remove a need for running everything manually.
The new script moves the `build.sh` functionality to Windows. I'm a bit reluctant to extend it too much with additional checks etc. because I think it should basically mirror what the bash.sh does.

GameTec-live:
Fyi; There are easy ish ways to run bash under windows;
namely Git bash and MSYS. Most people would probably build using docker anyways.
Doesnt hurt to have tho ig?

misch2:
OK, that docker build etc. makes perfect sense. Closing the PR.

LupusE:
This is what I wanted to say in my first comment. We don't need another way of build script. But If it would provide more features, it may is very useful. Maybe for the folks who don't want or can't use docker.
I could test it earliest next monday and I would give it a shot.

PR #287, opened 2025-08-29
Add LF Viking support

jamisonderek:
This PR adds support for the Viking protocol, which is a low-frequency protocol (RF/32 with Manchester encoding, 2 data blocks : 64-bit data). The card ID is an 8-digit hexadecimal number.

Place a T5577 against CU then run the following commands...

lf viking write --id 11223344
lf viking read
Next try emulation...

hw slot type -t Viking -s 4
hw slot enable -s 4 --lf
lf viking econfig --slot 4 --id 13371337
hw slot store
hw slot list
Notice the slot 4 will show the information. Select slot 4 and read with Viking reader (or Flipper Zero).
Try cloning a card using, Long press A while holding a Viking card (or programmed T5577) against CU & notice all LEDs blink green. If not, try Long press A again. Read CU emulation with a Viking reader (or Flipper Zero).

hw slot list should now show a nickname of cloned VIKING.

### DIFF
--- a/firmware/application/Makefile
+++ b/firmware/application/Makefile
@@ -36,6 +36,7 @@ SRC_FILES += \
   $(PROJ_DIR)/rfid/nfctag/lf/utils/manchester.c \
   $(PROJ_DIR)/rfid/nfctag/lf/protocols/em410x.c \
   $(PROJ_DIR)/rfid/nfctag/lf/protocols/hidprox.c \
+  $(PROJ_DIR)/rfid/nfctag/lf/protocols/viking.c \
   $(PROJ_DIR)/rfid/nfctag/lf/protocols/wiegand.c \
   $(PROJ_DIR)/utils/dataframe.c \
   $(PROJ_DIR)/utils/delayed_reset.c \
@@ -343,6 +344,7 @@ ifeq	(${CURRENT_DEVICE_TYPE}, ${CHAMELEON_ULTRA})
     $(PROJ_DIR)/rfid/reader/lf/lf_reader_main.c \
     $(PROJ_DIR)/rfid/reader/lf/lf_t55xx_data.c \
     $(PROJ_DIR)/rfid/reader/lf/lf_hidprox_data.c \
+    $(PROJ_DIR)/rfid/reader/lf/lf_viking_data.c \
 
   INC_FOLDERS +=\
     ${PROJ_DIR}/rfid/reader/ \
--- a/firmware/application/src/app_cmd.c
+++ b/firmware/application/src/app_cmd.c
@@ -650,6 +650,30 @@ static data_frame_tx_t *cmd_processor_hidprox_scan(uint16_t cmd, uint16_t status
     }
     return data_frame_make(cmd, STATUS_LF_TAG_OK, sizeof(card_data), card_data);
 }
+
+static data_frame_tx_t *cmd_processor_viking_scan(uint16_t cmd, uint16_t status, uint16_t length, uint8_t *data) {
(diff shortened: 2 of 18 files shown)


LupusE (approved):
As I tested the old PR already in detail, I've downloaded the update, compiled it and have done only what is written in the PR. works perfect against the Flipper as reader and emulator.

The position at the flipper is still al little time consuming, this time I've got the best results near the middle of the flipper. But this is already addressed at the latest LF PR.

Great work!

LupusE:
Regarding the compiling error:
> [Lint (pyrefly + ruff) / lint (pull_request)](https://github.com/RfidResearchGroup/ChameleonUltra/actions/runs/17327358491/job/49263941334?pr=287)Failing after 8s

Last weekend I fixed the ruff error by replacing `except:` with `except Exception:`.
It is okay for ruff, but needs testing for prod. I will prepare a branch for testing some ruff/pyrefly messages this week.

nfcmd:
@jamisonderek any hope for a fix on low frequency HID prox? Emulation still seems hit or miss. #325 for one.

jamisonderek:
> @jamisonderek any hope for a fix on low frequency HID prox? Emulation still seems hit or miss. #325 for one.

For HID prox, I would say @TeCHiScy is the best person to help investigate. If you are having issues with Viking protocol, I can take a look (but I would need to know the ID that you are emulating). In general for Viking, there were some issues with fully reading the signal, where the code doesn't invoke the interrupt on every signal transition - so it does a hack to try to figure out the timing of the missing signals. The timing hack worked in my testing for the Viking cards, but different IDs may expose a case where the missing signal is at a different spot than what I observed. For emulating, the Flipper Zero always read the Viking ID.

I'm still dealing with housing remodels, so I haven't done any Flipper or CU work for the past 6 months (I've been silent on YouTube as well). I hope mid-February that the housing projects are done and I'm able to jump back in & setup my Proxmark to see signals, etc. and help with more protocols for the CU (maybe one day I'll understand the FSK2a protocols, like H10301).

TeCHiScy:
> > @jamisonderek any hope for a fix on low frequency HID prox? Emulation still seems hit or miss. #325 for one.
> 
> For HID prox, I would say @TeCHiScy is the best person to help investigate. If you are having issues with Viking protocol, I can take a look (but I would need to know the ID that you are emulating). In general for Viking, there were some issues with fully reading the signal, where the code doesn't invoke the interrupt on every signal transition - so it does a hack to try to figure out the timing of the missing signals. The timing hack worked in my testing for the Viking cards, but different IDs may expose a case where the missing signal is at a different spot than what I observed. For emulating, the Flipper Zero always read the Viking ID.
> 
> I'm still dealing with housing remodels, so I haven't done any Flipper or CU work for the past 6 months (I've been silent on YouTube as well). I hope mid-February that the housing projects are done and I'm able to jump back in & setup my Proxmark to see signals, etc. and help with more protocols for the CU (maybe one day I'll understand the FSK2a protocols, like H10301).

Replied in another thread. Looking forward to see you back! Call me when you back into the CU world!

For me, to be frank, I've also busy at some other intresting projects these days. Also, the emulation of HID seems to related to signal strength, which needs an oscilloscope to dig into it.

nieldk:
viking_scan() in chameleon_cmd.py has no @expect_response decorator, so it returns a raw Response object. The CLI then does id = self.cmd.viking_scan() followed by id.hex(), which calls .hex() on a Response object, not on bytes — this will throw an AttributeError. Every other scan command either uses @expect_response so the decorator unwraps it to resp.parsed, or the CLI checks the status before using the result. Viking does neither.

PR #290, opened 2025-09-02
Clang format

GameTec-live:
Add clang-format as a formatter and linter to the firmware code

Oh and also run the formatter lol (resulting in the large diffs)


Foxushka:
Maybe merge https://github.com/RfidResearchGroup/ChameleonUltra/pull/274 before reformatting whole codebase?
And there is working ruff config https://github.com/Foxushka/ChameleonUltra/blob/d8b9bc310f9301c2496f3518ad2c0937dccc329a/ruff.toml

GameTec-live:
Good point. However that doesnt explain why thatd break builds lol
Ruff config is in the works, so that check can just be ignored for now. @LupusE is also working on fixing the pyrefly issues. (ignoring E722 isnt a solution lol)

Foxushka:
> (ignoring E722 isnt a solution lol)

Replacing `except:` with `except Exception:` still makes zero sense, as we just ignore all errors

GameTec-live:
> > (ignoring E722 isnt a solution lol)
> 
> Replacing `except:` with `except Exception:` still makes zero sense, as we just ignore all errors

yeah, but ig it helps with readability, but ideally there's proper error handling (another thing afaik @LupusE is atleast partially working on)

Anyways, that still WIP... the issue rn is to figure out why clang format breaks the builds...

LupusE:
> Replacing `except:` with `except Exception:` still makes zero sense, as we just ignore all errors

100% correct.
But we've got nearly 200 pyrefly errors per file and roundabout 80 rust.
At some point I just want to make the list shorter. And this replace worked. The alternative is to take look at every except and write a fitting handling.

I doubt this will be possible by anyone in reasonable time.
There will a lot of 'solutions' follow, that makes the linter quiet and only improves the readability, not the code quality.

I thought at this amount of errors, I can detect more pattern. But in reality, I was wrong.
WIP

PR #294, opened 2025-09-07
Add option for NFC Field Generator

RocketGod-git:
Added NFC Field Generator to the options in button settings.
Also added the option to Python CLI to set the option.
TODO: GUI integration.

https://github.com/user-attachments/assets/1ff5dc28-2ab0-4f1c-bad1-8eefcfa25e46

### DIFF
--- a/firmware/application/src/app_cmd.c
+++ b/firmware/application/src/app_cmd.c
@@ -488,6 +488,35 @@ static data_frame_tx_t *cmd_processor_mf1_write_one_block(uint16_t cmd, uint16_t
     return data_frame_make(cmd, status, 0, NULL);
 }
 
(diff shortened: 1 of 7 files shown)


LupusE (approved):
It's optional, and does not touch the card formats.
Fine for me.

I agree to @GameTec-live, it may be useful to have this also as CLI switch. But I have no time at the moment to implement. More important things needs to be done at the source.

RocketGod-git:
Can do that too... this is more to show off my implant LED on the go... would be good to add to cli also and control the field.

RocketGod-git:
If I add to CLI, it will be a separate PR since it's a whole different project to start.
This PR can be merged. There is nothing left to add.

PR #295, opened 2025-09-11
Use a function for colored strings

mischif:
Instead of manually setting ANSI colors everywhere, have a function handle it.

### DIFF
--- a/software/script/chameleon_com.py
+++ b/software/script/chameleon_com.py
@@ -4,7 +4,7 @@ import threading
 import time
 import serial
 from typing import Union
-from chameleon_utils import CR, CG, CC, CY, C0, color_string
+from chameleon_utils import CR, CG, CC, CY, color_string
 from chameleon_enum import Command, Status
 
 # each thread is waiting for its data for 100 ms before looping again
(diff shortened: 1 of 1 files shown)


LupusE:
Looks good. But I have some remarks.

From your function, it will set every string to C0 at the end. Looks clean to me.
But why is C0 active as import?
When I take a look at https://github.com/RfidResearchGroup/ChameleonUltra/blob/d440d98d121aed088503d3861cddad526f1b521f/software/script/chameleon_com.py#L222C37-L222C113 there is:
```
print(f"<={color_string((CC, command_string.ljust(40)), (C0, status_string),
```

When I read your new function, it should already terminate with C0, so it is not necessary to format the following part, isn't it?
Even if it am overlooking something, wouldn't it be more readable to format it also with your new function?
```
print(f"<={color_string((CC, command_string.ljust(40)), color_string((C0, status_string)),
```

If we are about to change the code, maybe we also should rename the variables:
```
c_red = colorama.Fore.RED
c_green = colorama.Fore.GREEN
c_blue = colorama.Fore.BLUE
c_cyan = colorama.Fore.CYAN
c_yellow = colorama.Fore.YELLOW
c_magenta = colorama.Fore.MAGENTA
c_reset = colorama.Style.RESET_ALL
```

In my opinion the short form has no advantage above the clear name and everyone knows exactly what is set. Opinions on this are welcome.

mischif:
It seems I got confused and thought that `C0` reference was meant to reset the following text completely and I didn't remember it gets converted to red anyway; fixed.

PR #303, opened 2025-09-30
Nix flake + python package

machshev:
This is probably too much in one PR, but if your interesting in going in this direction I could perhaps split this up into more incremental changes.

- [x] Add nix flake + direnv so that it's easier to develop on NixOS
- [x] Moved the Python cli code from the scripts dir into a python package structure. This means it could be published as a python package.
- [x] Ruff format - auto format to fix up some of the main lint issues and git a bit more consistency.


GameTec-live:
Hey, thanks for the PR!
A few things:
- Can we get a few more details on what got changed as its a pretty big PR
- Can you fix the autobuilds as it seems like your PR breaks the pyinstaller autobuilds among other things
and - Am I right in assuming, that these changes still keep the CLI standalone runnable as detailed in the current docs?

machshev:
The nix flake + direnv is a standalone change that I can pull out into a separate PR.

The python refactoring should just be:
- [ ] creating a new directory for the python source - which becomes the top level python package name (hence chameleon). So scripts is not a great namespace, it's normally the same or close to the python wheel name. Quite often there would be a src dir and then the package name. But with src already containing C code that's not practical.
- [ ] update the pyproject.toml to
  - [ ] tell it where to find the python source as it's not in a standard place now.
  - [ ] move the dev dependencies def to the standard location so UV and pip  understands them. It's then possible to do `pip install -e .[dev]` for example. 
  - [ ] simplify the package name in case you wanted to release to pypi at some point.
  - [ ] add a script entry point so the tool is available on PATH when the python package is installed. 
- [ ] add `__init__.py` which is a marker file to tell the interpreter that this directory is a python package.
- [ ] rename the python modules to remove the chameleon namespace in the filename. The namespace is there already now with the name of the parent directory "chameleon" being part of the module path.
- [ ] fix up the imports making them fully qualified module path imports. Preferring the recommended explicit `from ... import ...` so that only the parts of the module required are imported. 

I think that's it in terms of changes. The tests are still in the scripts dir, but would normally be in a `tests` dir in the same dir as the `pyproject.toml`. I'm not sure what pyinstaller is for, I'll look it up later. 

If this is a direction you want to go i.e. a more standard python package that could be installed with `pip install chameleon` (if you uploaded a package to pypi). Then I'll try and find some time to tidy this up.

The way to run the cli is more like a standard installed program than a "script". Although it would be easy to put back a wrapper script to fix up the docs... or I could fix up the docs to match?

GameTec-live:
Thanks!
Ill need to pull and test this PR to see how its currently behaving, but the important part is that one can clone the repo, install deps and just run the CLI without needing to compile or install anything (ok, C parts excluded), etc.

Feel free to pull the Nix changes into a separate PR, as that I think I can just merge as its only introducing a new packaging variant / config file. Dont have to tho if you dont want to.

felurx:
I think it would be very nice to have the CLI as a python package. Once that's done, it should be fairly easy to make the CLI installable with pipx. That'd be pretty convenient, you just run `pipx install ...` and pipx creates the venv, installs dependencies, and makes sure the program is on your PATH. 

As far as I can see from [the pipx docs](https://pipx.pypa.io/latest/how-pipx-works/#developing-for-pipx), this should be just a few lines in `pyproject.toml`.
After the python package stuff is done and this is merged, I'd take a look at that and submit a PR if I'm successful :)

GameTec-live:
@felurx might be worth targeting UV instead, but as its just a python package it should work on both.
https://docs.astral.sh/uv/#tools

GameTec-live:

felurx:
What would still need to be done? Is it all the things from https://github.com/RfidResearchGroup/ChameleonUltra/pull/303#issuecomment-3359580849?
I still think it'd be nice to have the Python package, but I probably can't put much effort into it currently.

GameTec-live:
@felurx This PR is simply out of date and hasnt kept up with the rest of the development. (see merge conflicts)

It also never was marked ready for review.

PR #306, opened 2025-10-08
Adds generic ADC read functionality

merlokk:
This introduces a new command to sample the ADC values from the LF antenna and returns them to the user.


LupusE (approved):
To me it more looks like an debugging information. But I at this point maybe helpful to debug some non working readers?

PR #308, opened 2025-10-16
Fix chameleon_cli_unit.check_tools check the wrong dir

azuwis:
Commit 35d2f40ff5801812c3517be89a4e1ad9c24b95ad introduce a breaking change, `chameleon_cli_unit.check_tools` check `bin` dir from pwd, instead of `bin` dir in the same dir of `chameleon_cli_unit.py`.

`python3 ./software/script/chameleon_cli_main.py` outputs wrong warning:

```
Warning, ... not found. Corresponding commands will not work as intended
```

### DIFF
--- a/software/script/chameleon_cli_unit.py
+++ b/software/script/chameleon_cli_unit.py
@@ -65,12 +65,11 @@ def load_dic_file(import_dic, keys):
 
 
 def check_tools():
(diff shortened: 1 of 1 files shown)


azuwis:
Close in favor of #317

PR #313, opened 2025-10-24
Fix random crash for >1024 bytes in data_frame_make

Foxushka:
Reported in https://github.com/GameTec-live/ChameleonUltraGUI/issues/753

I don't really like this solution, but it is reliable. Maybe we should disable logging in "release" builds (if this is possible with NRF)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Fix random crash for >1024 bytes in data_frame_make (@Foxushka)
  - Fix for static nested key recovery (@jekkos)
 
 ## [v2.1.0][2025-09-02]


suut:
I spent quite some time tracking down this bug, I narrowed it down to an issue with the nRF logging frontend and in particular potentially in `nrf_log_frontend_dequeue`
It doesn't look particularly easy to fix so I think we should merge this PR until we get a more proper fix (if we find one)

Disabling logging in release builds or at least if the debugger is not attached is a good idea as well

GameTec-live:
@suut You sure this didnt get fixed with #299 ? Thought that includes this fix too?

suut:
@GameTec-live this other PR also fixed it by commenting out the code that was causing the bug, but we might still want to have the logging back

Maybe @xianglin1998 can chime in; my plan was to have a dedicated RTT buffer for the dataframe (and other hexdumps) so it doesn't show in the main RTT terminal but you can still access the data if you want, and it bypasses nrf_log so it doesn't crash anymore, what do you think of that? I have a working proof of concept for it

suut:
basically this, maybe with a little macro to add some text around the hexdump; this is all lock-free, not to be called from an ISR, and it the buffer is full (if SWD isn't connected and nobody is reading from the buffer) it won't hang
```c
#define RAW_DATA_PER_BLOCK ((size_t)128)

__attribute__((__aligned__(16))) static uint8_t bulk_data_buffer[16384];
static char bulk_data_fmt_buf[3 * RAW_DATA_PER_BLOCK]; // can fit 128 data bytes after formatting
static char bulk_data_hex_lut[16] = {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F'};
int bulk_data_buffer_index = -1;

static inline void log_bulk_data(const uint8_t *const data, size_t size) {
    if (bulk_data_buffer_index < 0 || !size || ((__get_IPSR() & IPSR_ISR_Msk) >> IPSR_ISR_Pos) != 0) {
        // We exit if we are in an interrupt handler or the data is empty or the buffer isn't allocated yet
        return;
    }

    const size_t size_formatted_data = 3 * size; // 3 bytes per input byte: 2 hexadecimal digits and a space/newline
    const size_t num_output_blocks = 1 + (size_formatted_data - 1) / sizeof (bulk_data_fmt_buf);
    

    for (uint32_t block = 0; block < num_output_blocks; block++) {
        const bool last = (block + 1) * RAW_DATA_PER_BLOCK > size;
        const uint32_t bound = last ? size % RAW_DATA_PER_BLOCK : RAW_DATA_PER_BLOCK;
        for (uint32_t i = 0; i < bound; i++) {
            const uint8_t rd = data[block * RAW_DATA_PER_BLOCK + i];
            char *const wr = &bulk_data_fmt_buf[3*i];
            wr[0] = bulk_data_hex_lut[rd >> 4];
            wr[1] = bulk_data_hex_lut[rd & 0x0F];
            wr[2] = (i == bound - 1 && last) || (i > 0 && i % 16 == 0) ? '\n' : ' ';
        }
        SEGGER_RTT_WriteSkipNoLock(bulk_data_buffer_index, bulk_data_fmt_buf, 3 * bound);
    }
}


/**@brief Application main function.
 */
int main(void) {
    log_init();               // Log initialization
    // Initialize the bulk data logging after nrf_log initialized RTT
    bulk_data_buffer_index = SEGGER_RTT_AllocUpBuffer("Raw data", bulk_data_buffer, sizeof (bulk_data_buffer), SEGGER_RTT_MODE_NO_BLOCK_SKIP);
    // ...
}
```

Then you get something like this in a log file when you start the data logging in the Segger RTT viewer (I haven't tested with openocd yet):
```
# SEGGER J-Link RTT Viewer V8.76 Data Log File
# Compiled: 11:42:30 on Oct  8 2025
# Logging started @ 26 Jan 2026 02:25:27
72 6F 66 6C 6D 61 6F 6D 64 72 6C 6F 6C 70 74 64
72 68 65 6C 6C 6F
```

GameTec-live:
@suut so rn it is fixed, even if not nicely and you plan on working on a proper solution?

Id just close this PR then and wait for your proper solution?

suut:
Sure!

PR #316, opened 2025-10-30
Fix HIDProx formatting with `hw slot list`

brewt:
Fix HIDProx formatting to match the formatting of all other card types.


LupusE (approved):
We have a PR for changing the colour handling. But I think this is already merged. So go for it.

PR #317, opened 2025-11-01
fix: resolve binary tools detection after PyInstaller packaging

bob-zebedy:
This fixes the warning about missing tools after PyInstaller

packaging by correctly detecting the temporary

extraction directory (sys._MEIPASS) in frozen environments.

- New get_resource_dir() in chameleon_utils.py

- Import default_cwd in chameleon_cli_unit.py

- Simplify check_tools() to use imported default_cwd


GameTec-live (approved):
LGTM

LupusE (approved):
Does not break other components. I have not tested the pyinstaller myself, but looks reasonable.

PR #332, opened 2026-01-19
fix hidprox UnboundLocalError cli arg error + hidprox slot set warning

RickConsole:
When manually configuring emulation settings for hidprox cards, there is an UnboundLocalError which is caused by not assigning local variables (fc, cn, il, oem) whne the print statements attempted to use them. 

```
[USB] chameleon --> lf hid prox econfig -f H10301 --fc 123 --cn 45678 -s 1
 - Set hidprox tag id success.
CLI exception: Traceback (most recent call last):
  File "/home/rick/src/ChameleonUltra/software/script/chameleon_cli_main.py", line 146, in exec_cmd
    raise error
  File "/home/rick/src/ChameleonUltra/software/script/chameleon_cli_main.py", line 141, in exec_cmd
    unit.on_exec(args_parse_result)
  File "/home/rick/src/ChameleonUltra/software/script/chameleon_cli_unit.py", line 3236, in on_exec
    if fc > 0:
       ^^
UnboundLocalError: cannot access local variable 'fc' where it is not associated with a value
```

Additionally, added a warning when setting a HIDProx card to a slot that isnt configured to be one (Similar to Viking's functionality)


LupusE (approved):
The CLI message is a great addition! Very helpful for the users.

PR #335, opened 2026-01-20
Ultimate Magic Card support in the CLI

suut:


software/script/chameleon_utils.py
```diff
@@ -54,8 +54,7 @@ def __init__(self, *args, **kwargs):
         self.help_requested = False
 
     def exit(self, status: int = 0, message: Union[str, None] = None):
-        if message:
-            raise ParserExitIntercept(message)
+        raise ParserExitIntercept(message)
```
suut: This fixes the bug it would show twice the `--help` of some commands

LupusE (approved):
I don't own an 'Ultimate Magic Card' (Gen4). But the worst what could happen from reviewing the code: we have a card that should work, but doesn't ... 
Maybe there will be a merge conflict with the double-help fix. This should be an own PR

suut:
Thanks, I need to revisit this a bit and add the rest of the missing code; I have a new version UMC to test with but not the old version; but the differences are minimal and already taken into account by the script 

I'll split into another PR the help fix

LupusE:
Thanks a lot for your update.
For the double-help fix: I just saw @nieldk mentioned it a couple days ago with a fix. If you both coordinate the PRs, it should be fine. We just need to keep an eye on this.
If we reduce 2 help messages 2 times -> 2-(2x1) = 0 ... We have no help in the end. Pure math :P

nieldk:
T55xx clone by nieldk · Pull Request #401 · RfidResearchGroup/ChameleonUltragithub.comSent from my iPhoneOn 12 Apr 2026, at 19.21, Benjamin Møller ***@***.***> wrote:﻿LupusE left a comment (RfidResearchGroup/ChameleonUltra#335)
Thanks a lot for your update.
For the double-help fix: I just saw @nieldk mentioned it a couple days ago with a fix. If you both coordinate the PRs, it should be fine. We just need to keep an eye on this.
If we reduce 2 help messages 2 times -> 2-(2x1) = 0 ... We have no help in the end. Pure math :P

GameTec-live:

PR #337, opened 2026-01-24
TCP transport in the CLI to support Android (Termux)

suut:
This adds support for building on Android in Termux in order to use the CLI through a USB-serial TCP bridge


suut:
Note: the static type checker and the Windows builds fail, but it doesn't seem related to the code I added; I fixed the build for MacOS and the linter checks

PR #341, opened 2026-01-26
Fix typo in hid prox econfig and add ACTProx HID card type

MusicLeecher:
- fix typo in hidprox econfig set/get function which would generate an "UnboundLocalError" when using lf econfig to set a new card

### DIFF
--- a/firmware/application/src/rfid/nfctag/lf/protocols/wiegand.c
+++ b/firmware/application/src/rfid/nfctag/lf/protocols/wiegand.c
@@ -553,7 +553,7 @@ static uint64_t pack_actprox(wiegand_card_t *card) {
     if (card->oem == 0) {
         card->oem = 900;
     }
-    uint64_t bits = PREAMBLE_36BIT;
+    uint64_t bits = PREAMBLE_ACTP;
     bits <<= 1;
     bits = (bits << 10) | (card->oem & 0x3ff);
     bits = (bits << 8) | (card->facility_code & 0xff);


LupusE:
The change seems to more as described in the comment.
How can the new card type be tested? Is there a reference reader or something that could emulate this to test?
I can imagine Wiegand should be easy to emulate on a Proxmark3?

GameTec-live:
I noticed that it looks like a new cardtype but I thought that that's just the most effective way to solve the issue as that's maybe some weird permutation of HID Prox?

GameTec-live:
Therefore someone who had that error before should probably test if this new type solves it?
But yeah, this is more than a typo

MusicLeecher:
> The change seems to more as described in the comment. How can the new card type be tested? Is there a reference reader or something that could emulate this to test? I can imagine Wiegand should be easy to emulate on a Proxmark3?

I made the PR after I compiled the firmware and it's the version I've been using on my chameleon, although I understand this requires further testing because AFAIK the decoding of the UID, FC, OEM could be all wrong and the only solution to that would be to get a hold of a ACTProx tag with known information and see how it compares to the decoded values on the Chameleon... As far as a cloning/emulation goes, it's been working fine for me and before being in use everyday the raw values from the tag I have and the Chameleon output were checked with a proxmark3.

MusicLeecher:
> Therefore someone who had that error before should probably test if this new type solves it? But yeah, this is more than a typo

As for the error that's been fixed it's easy to check what is happening on the current build you can try this: `lf hid prox econfig --cn 0000000000 --fc 0 --format HGEN37`
You will seen the error I mentioned and the fix was just to indent the functions to their right place; it's the changes made in chameleon_cli_unit.py form line 3688 to 3694

MusicLeecher:
> I noticed that it looks like a new cardtype but I thought that that's just the most effective way to solve the issue as that's maybe some weird permutation of HID Prox?

I decided to create a new card type as I was not 100% sure it's just a permutation of HID Prox, there are more efficient ways to do it if indeed we conclude it's just the preamble that changes from the ACT Pro tags to the HID Prox tags; but if there are more changes made to the decoding/encoding of the ACT Prox tags then having a separate card type will make changes easier in the future, I think.

MusicLeecher:
#332 seems to also be a fix to the same error I mentioned in my PR

MusicLeecher:
> Approved by me, although I can't test it on a 'real' device, it does not break anything else.
> 
> @MusicLeecher Could we close #332 with this approval as well? Would it be okay for you?

yes I would approve and merge #332 as well

PR #343, opened 2026-02-02
Fix compiler error  `a label can only be part of a statement and a de…

LupusE:
…claration is not a statement`

It's only a compiler error, regarding to https://www.educative.io/answers/resolving-the-a-label-can-only-be-part-of-a-statement-error

The change was coming with https://github.com/RfidResearchGroup/ChameleonUltra/pull/314/changes#diff-8e81c0cb796bfec7b1cd15724c4adbeb5bc55bf89909f833de73d29e1ef4a608 and I have no Idea how it compiled before.

### DIFF
--- a/firmware/application/src/rfid/nfctag/hf/nfc_mf0_ntag.c
+++ b/firmware/application/src/rfid/nfctag/hf/nfc_mf0_ntag.c
@@ -704,11 +704,12 @@ static bool check_ro_lock_on_page(int block_num) {
         switch (m_tag_type) {
             case TAG_TYPE_NTAG_213:
             case TAG_TYPE_NTAG_215:
-            case TAG_TYPE_NTAG_216: ;
+            case TAG_TYPE_NTAG_216: {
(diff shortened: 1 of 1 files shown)


suut:
You can also enclose the contents of the case within `{ }`, it silences the warning but also has the added effect of not defining the variable globally so you don't end up in that case:

```
switch (2) {
  case 1:
    int x = 2;
    break;
  case 2: ;
    printf("%d\n", x);
    break;
}
```

Here it would print a random value for `x`, it's declared globally because it's part of the switch case in the same scope, but it's not defined because case `1` was not executed, so you end up with a random value.

You have the same kind of problem if you declare variables with the same name in two different cases, they would clash together 

So instead you can do
```
switch (2) {
  case 1: {
    int x = 2;
    break;
  }
  case 2: ;
    printf("%d\n", x);
    break;
}
```

LupusE:
I changed the fix as suggested with curly braces instead of column, to unify the code.

PR #346, opened 2026-02-03
fix cli arg parser for FIELD_OFF_DO_RESET

YuyangisCoding:
Fixed cli arg parser error introduced in PR#342


xianglin1998:
Merged, good!

PR #352, opened 2026-02-06
Fix bad missing tools warning

suut:


LupusE (approved):
Just an working directory change, seems to work, no harm.

PR #353, opened 2026-02-07
API Auto workflow

gokingsgo805:
IN THE CLI, ALL THAT IS REQUIRED IS THE hw_connect COMMAND
API INTGEGRATION THEN INITIATES FIRST, LF SCAN, THEN IT DETERMINES IF IT IS AN lf OR hf.  READS BLOCK DATA, APPLIES KEYS, AND DUMPS THE RESULTS.
IF hf, READS, DETERMINES THE CHIP TYPE, DETERMINES KEY DICTIONARY AND PERFORMS ATTACKS, READS VALUE, IN HEX & OUPUTS IN DECIMAL

### DIFF
--- a/software/script/chameleon_cli_unit.py
+++ b/software/script/chameleon_cli_unit.py
@@ -682,8 +682,8 @@ class AutoWorkflowRunner:
         try:
             if not self._ensure_reader_mode():
                 return
-            hf_handled = self._handle_hf()
             lf_handled = self._handle_lf()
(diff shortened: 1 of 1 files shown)


gokingsgo805:
I AM NEW.  DONT CRUCIFY ME.  This is my first attempt at coding anything except my 3D printer board.
I have little idea as to what im doing.  i just thought it would be cool to not have to figure out CLI commands or using the gui

PR #358, opened 2026-02-15
feat: add launcher script for CLI quick start

fmuk:
## Summary

- Add `chameleon.sh` launcher script that auto-creates a Python venv and installs dependencies on first run
- Add `_archive/` to `.gitignore`

### `chameleon.sh`

A convenience launcher for new users. On first run it:
1. Checks for Python 3.9+
2. Creates a venv in `software/.venv`
3. Installs dependencies from `requirements.txt`
4. Launches the CLI

On subsequent runs it skips setup and goes straight to the CLI. Works from any directory.

```bash
./chameleon.sh
```

## Test plan

- [ ] First run: verify venv creation and dependency installation
- [ ] Second run: verify it skips setup and launches directly
- [ ] Run from a different directory: verify it resolves paths correctly
- [ ] Run without Python 3.9+: verify it shows a helpful error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -708,3 +708,6 @@ FodyWeavers.xsd
 # End of https://www.toptal.com/developers/gitignore/api/visualstudio,c++,c,python,visualstudiocode,macos,windows
 software/script/tests/nonces.bin
 software/script/nonces.bin
+
+# Local archives / investigation artifacts
+_archive/


GameTec-live:
Did you give this PR any thought or did you blindly commit claudes output?

This approach is neither especially clean nor sensical. Various launch and build options already exist and hardcoding the dependencies is a big nono.

As a bonus; with the exception of compiling the binaries, uv already handles everything for you. (venv, deps, etc)

PR #386, opened 2026-03-21
add EM4X05 (read)support for CU

nieldk:


nieldk:
**Testing**: The CLI command lf em 4x05 read is accessible and returns _LF_TAG_NO_FOUND_ gracefully when no tag is present. Full protocol testing requires a physical EM4x05 or EM4x69 tag — the RTF gap-command protocol cannot be simulated with lf simask as it requires the simulator to respond to reader commands rather than broadcasting continuously. Hardware verification can be performed by a maintainer with access to EM4x05 tags (common in animal microchip readers, some parking fobs, and industrial access control systems).

PR #387, opened 2026-03-21
Support for changing the wake time in the client

### LINKED ISSUE #242: [idea] Support for changing the wake time of Chameleon in the official command line/Chameleon GUI
Please keep in mind we are still at the infancy of the project and a lot has still to be done.

Have first a look at the [Whitepaper](https://github.com/RfidResearchGroup/ChameleonUltra/blob/main/docs/technical_whitepaper.md), the Wiki [Roadmap](https://github.com/RfidResearchGroup/ChameleonUltra/wiki/Public-Roadmap) and [Wishlist](https://github.com/RfidResearchGroup/ChameleonUltra/wiki/Wishlist).  
No need for creating issues for well known missing supports, we are already aware :)

**Is your feature request related to a problem? Please describe.**
As stated in the question

**Describe the solution you'd like**
The Chameleon's wake duration can be modified from the official command line via a command like hw device sleep +time(ms) or from the device Settings page in the Chameleon GUI, for example: hw device sleep 8000 means that the wake-up time of chameleon is modified to 8s, and the duration verification is required. If the duration exceeds 30s(30,00ms), a warning will be issued, but the user can still set such a long wake-up time by "making sure he understands the risk". If the delay is longer than 60 seconds (60,00ms) or less than 1 second (1000ms), an error will be thrown to deny the user's request. Too long wake-up time will cause the chameleon to "brick" without a computer because there is no direct command to turn the chameleon off (sleep), and too short wake-up time will cause the user to operate the chameleon on mobile devices (Android,iOS) without sufficient time to complete the Bluetooth connection to the device.

**Describe alternatives you've considered**
You can set the long-hold function of the A/B key to wake up (30s), and hold it again to cancel wake up (that is, hold the wake up only for a fixed time in the firmware, 8s in the 2.0.0 firmware).

**Additional context**
NULL ：)

naaraxi:
fixes #242


naaraxi:
0d14309 to b026f29 - quick fix to appease ruff
there are some other preexisting linter issues but they don't really seem to break anything

GameTec-live (approved):
looks good, will test once I have time

naaraxi:
should i rename the setting name? maybe "sleeptimeout" is a bit vague..

naaraxi:
Conflicts resolved

LupusE (approved):
Tested the function, not anything else. Looks okay from the source.

```
[USB] chameleon --> hw settings sleeptimeout -s 5
Wake timeout set to 5 seconds.
Do not forget to store your settings in flash!
[USB] chameleon --> hw settings store
Storing settings...
 - Store success @.@~
```

Approve from my side.

LupusE:
Nice work. I think I can't appreciate i right now as I should ... But sometimes the fast-sleep was really annoying.
Thanks for the contribution @naaraxi

PR #388, opened 2026-03-23
Fix firmware application USB serial number

taichunmin:
Fix the USB serial number to distinguish difference devices.

Following is the port info before fix:

```js
[
  { // application
    path: '/dev/tty.usbmodem0000000000001',
    manufacturer: 'Proxgrind',
    serialNumber: '000000000000',
    pnpId: undefined,
    locationId: '14110000',
    vendorId: '6868',
    productId: '8686'
  },
  { // bootloader
    path: '/dev/tty.usbmodemD859C13EE4AD1',
    manufacturer: 'Proxgrind',
    serialNumber: 'D859C13EE4AD',
    pnpId: undefined,
    locationId: '14110000',
    vendorId: '1915',
    productId: '521f'
  }
]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Fix firmware application USB serial number (@taichunmin)
  - Added ioProx LF protocol support (read, emulate and T55xx clone)
  - Added commands to dump and clone Mifare tags
  - Fix bad missing tools warning (@suut)
--- a/firmware/application/src/sdk_config.h
+++ b/firmware/application/src/sdk_config.h
@@ -6614,7 +6614,7 @@
 
 
 #ifndef APP_USBD_STRING_SERIAL_EXTERN
-#define APP_USBD_STRING_SERIAL_EXTERN 0
+#define APP_USBD_STRING_SERIAL_EXTERN 1
 #endif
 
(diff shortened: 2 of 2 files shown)


LupusE:
I understand it will define the identification in a better way but what exactly?
Does it affect clones? Would we lock them out? I have no clone, so I can't try if would have an effect.

taichunmin:
When I connect 2 devices to macOS, the following screenshot is the devices list of ChameleonUltraGUI and chameleon-ultra.js:

| ChameleonUltraGUI | chameleon-ultra.js |
| --- | --- |
| Xnip_2026-03-29_234210 | Xnip_2026-03-29_234729 |

After I switch 2 devices into DFU mode, the following screenshot is the devices list of ChameleonUltraGUI and chameleon-ultra.js:

| ChameleonUltraGUI | chameleon-ultra.js |
| --- | --- |
| Xnip_2026-03-29_235026 | Xnip_2026-03-29_235005 |

Device names in DFU mode are significantly more easily recognizable, reducing user confusion when performing firmware updates via the GUI.

taichunmin:
This PR will not break clones because:
1. This change just follow the same code in bootloader [\[1\]](https://github.com/RfidResearchGroup/ChameleonUltra/blob/e5d615d512e3afa7791117bd4a892b0b2c8e3600/firmware/bootloader/src/sdk_config.h#L1590), [\[2\]](https://github.com/RfidResearchGroup/ChameleonUltra/blob/e5d615d512e3afa7791117bd4a892b0b2c8e3600/firmware/bootloader/src/sdk_config.h#L1598)
2. I used two clone devices for development.
PXL_20260329_164418689

LupusE (approved):
Fine for me. Use of the SN instead of a generic 0lin  is a pretty good idea

PR #402, opened 2026-04-09
Add LF PSK Indala support

kevihiiin:
Adding emulation, T55XX writing, and T55XX-based reading for the Indala 64-bit PSK1 protocol.

| Feature | Details |
|---------|---------|
| Frame | 64-bit PSK1, fc/2, RF/32 |
| Data | 8 bytes (raw 64-bit frame) |
| TAG_TYPE | 310 (`TAG_TYPE_INDALA`) |
| Emulation | Yes* -- CU emulates, verified via correlation-based PSK decoder |
| Reader | Yes* -- T55XX only, via fc/8 DDC |
| T55XX Write | Yes -- fc/2 default (real-reader compatible), fc/8 override for self-test |
| Commands | SET/GET 5026/5027, SCAN 3035, WRITE_TO_T55XX 3036 |

This is would be the first PSK protocol for the CU. I have also ported Keri, NexWatch, Motorola and IDTECK protocols for the CU, but they rely on the same files as Indala, and I thought we might as well test this one first.

New firmware components:
- **PSK1 modulator** in `lf_tag_em.c` -- Timer3+PPI+GPIOTE driven FET switching for jitter-free PSK phase control
- **PSK generic reader** (`lf_psk_reader.c`) -- SAADC-based reader using Digital Down-Conversion (DDC) at fc/8
- **PSK demod utility** (`pskdemod.c/h`) -- period-8 DDC mixer, differential PSK1 decode, cumulative XOR integration
- **Indala codec** (`indala.c/h`) -- encoder, decoder, modulator, T55XX writer
- **Python client + CLI** -- `TagSpecificType.Indala = 310`, set/get/scan/write commands

**TLDR**; The firmware in theory works, with some nifty tweaks. There are hardware limitations (envelope detector) that, I beleive, make it impossible to read Indala (and PSK in general) tags at the fc/2 subcarrier frequency. Emulation is also tricky, as we cannot sample the edges of the carrier wave at fc/2 (62.5kHz) frequency. 

---

### Emulation (PSK1 fc/2 RF/32)

The modulator uses the nRF52840's Timer3 + PPI + GPIOTE peripherals to produce PSK. Unlike ASK or FSK, PSK requires **phase inversion** of the subcarrier while maintaining constant amplitude.

The initial approach used the **standard PWM DMA playback path** (same `nrfx_pwm_simple_playback` used by all ASK/FSK protocols), encoding phase via the nRF52 PWM bit-15 polarity control (`0x8001` = inverted, `0x0001` = normal). This produced **periodic DMA jitter**. However, the nRF52's DMA engine shares the AHB bus with other peripherals, causing a timing glitch every ~8000 samples from bus contention. For ASK/FSK this is acceptable because small timing jitter only slightly affects amplitude or frequency. But for PSK, **phase IS the information**,  any timing jitter directly corrupts the carrier phase relationship the reader is trying to measure.

The solution replaces PWM DMA entirely with a **Timer3 ISR + PPI + GPIOTE** path. A hardware timer fires at exactly 125 kHz. PPI triggers GPIOTE SET/CLR tasks to toggle the FET with zero software involvement per carrier cycle, the ISR only runs once per bit (every 16 cycles) to load the next phase value from a precomputed pattern buffer. This eliminates DMA-related jitter entirely. Each bit gets 16 fc/2 subcarrier cycles. Phase encoding is achieved by swapping the FET ON/OFF order within each carrier cycle:
- **Bit 0**: FET ON first half, OFF second half
- **Bit 1**: FET OFF first half, ON second half (180° phase shift)

Several hardware-dependent parameters were tuned empirically for (my?) CU:
- **FET pulse width** (`PSK_PULSE_TICKS = 72`, i.e. 4.5 µs): Swept from 108 down to 72 ticks to find the LC tank resonance sweet spot where the antenna produces the strongest coupled signal. The resonance band sits at 4.5–5.25 µs (72–84 ticks). 72 ticks was chosen for cross-unit tolerance, a ±1 µs shift from component variation still stays within the band. This parameter is hardware-dependent and would need re-tuning on boards with different antenna characteristics.
- **PWM clock** (`counter_top = 4`, 500 kHz): The nRF52840 requires COUNTERTOP >= 3. At the default 125 kHz LF clock, counter_top would be 1 (invalid). Bumping to 500 kHz gives counter_top=4, providing sufficient resolution for the ON/OFF phase pattern.
- **Broadcast count** (`LF_PSK_BROADCAST_MAX = 30`): PSK protocols need a longer gap-free signal than ASK/FSK (which use 10 repeats). PM3 captures ~314 ms of signal; 30 repeats = ~491 ms ensures the capture window is entirely filled. Tested up to 100, counterproductive because PM3 uses an absolute error threshold, so more frames = more total errors counted.

### Reader (PSK1 fc/8 DDC)

The reader uses the standard 125 kHz SAADC sampling path (same hardware as ASK/FSK readers) with a custom Digital Down-Conversion decoder. After extensive investigation of fc/2 decoding approaches (see "Hardware limitations"), **fc/8 was chosen as the only viable subcarrier** for CU's analog frontend:

1. **fc/8 subcarrier** (15.625 kHz), T55XX tags configured for fc/8 instead of standard fc/2. This frequency passes cleanly through CU's envelope detector, producing ~300x stronger signal than fc/2
2. **Period-8 rectangular wave mixer**: `cos8 = [+1,+1,+1,+1,-1,-1,-1,-1]`, `sin8 = [-1,-1,+1,+1,+1,+1,-1,-1]`,extracts orthogonal I/Q components from 32 samples per bit. DC cancels naturally (both waves sum to zero)
3. **Alignment search**: Scans all 32 sample offsets within one bit period to find the correct bit boundary (maximizes total IQ magnitude)
4. **Differential PSK1 decode**: Dot product of adjacent IQ vectors, negative = phase transition, positive = same phase
5. **Cumulative XOR integration**: Recovers raw data from differential bits. Starting phase is unknown, so both the decoded register and its complement are checked against the preamble
6. **Indala preamble validation**: 33-bit pattern `{1,0,1, 29×0, 1}` checked via bitmask `0xA000000080000000`

Hardware-dependent reader parameters:
- **Settling skip** (`INDALA_SKIP = 1024` samples, ~32 bits): The first ~8 ms of SAADC data after radio start contains the carrier ramp-up transient. Skipping it prevents the alignment search from locking onto the transient instead of real data.
- **T55XX power-on delay** (10 ms): The T55XX tag needs time after the carrier starts to complete its power-on reset before it begins modulating. Without this delay, the SAADC calibrates against an unmodulated carrier.
- **Buffer size** (6144 samples, ~3 frames): Sized to hold enough data for reliable preamble search while staying within the 8 KB heap limit. All PSK protocols share a single static buffer to avoid RAM overflow.

### T55XX Write

Encodes 64-bit data into 3 T5577 blocks (1 config + 2 data). **Default config uses fc/2** (matching PM3 and real Indala readers). An optional `fc8_override` parameter rewrites the PSK CF bits to fc/8 for CU self-testing. This means tags written by CU are compatible with standard Indala readers out of the box.

---

### Emulation
- CU emulates → PM3 reads via correlation-based PSK decoder
- **8/8 tests pass**: 3 test vectors (PM3 example, FC:123/CN:1337, FC:1/CN:1), protocol identification, persistence across slot switch, persistence across reboot, multi-read consistency

### Reader T55XX
- CU writes to T55XX tag (fc/2 config), reads back with fc/8 override
- **5/5 tests pass**: write-and-read, all 3 vectors, overwrite, consistency, scan-to-slot roundtrip

### Build
- `make -C ChameleonUltra/firmware/application` compiles clean
- Preflight checks 10/10

---

### Emulation: PM3 native PSK demod (~500 errors vs threshold of 100)

CU's emulated PSK signal carries valid data but produces waveform artifacts that exceed PM3's strict `pskRawDemod_ext` error threshold. **16 firmware approaches** were tested to improve signal quality:

**Signal tuning:** HFXO crystal clock request, FET duty cycle sweep (108–127 ticks, 9 values), all 32 GPIO drive mode combinations, phase training enable/disable, none had measurable effect on PM3 error count.

**Timing compensation:** Fractional tick drift compensation (−0.10 to +0.10), proved clock drift is not the dominant error source. Increased broadcast count (30→100), counterproductive. Phase-hopped frame repeats, worse due to boundary discontinuities.

**Waveform shaping:** Pre-distortion pulse tables (context-dependent FET timing after transitions and long runs), active quench pulses via Timer3 CC[2]/CC[3] during FET-OFF half-cycles, ISR software Q-spoiling at ~2 MHz, intermediate load keeper cycles, entry waveform / preamble conditioner, inter-frame settle gaps (0–10 ms), fixed phase remap (inverting fc/2 polarity), high-transition payload testing, none improved the error count.

**Root cause:** CU uses **active resistive load modulation**, an FET shorts/opens the antenna coil. Real PSK tags use **passive carrier-locked backscatter**, phase-coherent with the reader's carrier. The active FET switching on the LC tank produces ringing artifacts (measured lag-1 autocorrelation = +0.32 vs real tag's −0.47). This is a fundamental property of the modulation topology, not fixable in firmware.

**Resolution:** A correlation-based PSK decoder (dot-product of each 32-sample bit window against a reference carrier) decodes **100% of CU's PSK captures**. The data IS in the signal,PM3's edge-based demod is too strict for the active waveform shape. PM3 `lf search` still identifies Indala ~50% of the time.

### Reader: fc/2 (62.5 kHz) invisible to CU hardware

The standard Indala subcarrier is fc/2 (62.5 kHz). Multiple approaches were tried across several sessions before determining this is a hardware limitation:

**At 125 kHz sampling (standard SAADC path):** Period-2 IQ correlation, fc/2 sits exactly at Nyquist, folds onto DC. Period-4 IQ correlation, signal present but amplitude depends on the PM3↔CU carrier phase offset (independent oscillators with ~6 Hz beat frequency), causing bits to fall in unpredictable dead zones.

**At 166.67 kHz oversampling (custom TIMER3+PPI path):** Direct TIMER3 register writes + PPI → SAADC, SAADC internal timer continuous mode, both produced zero samples. Notch filter to remove carrier alias at fs/4, remaining PSK sidebands attenuated ~30–40 dB by antenna Q.

**Interleaved IQ via dual TIMER3 PPI triggers:** Two samples per carrier cycle at 90° separation for I/Q extraction. Accumulated differential decoder, exhaustive alignment search, negative-R quality metric, local DC cancellation (D-metric), alignment discrimination between correct and off-by-1 was only ~1.8x, insufficient given the SNR.

**Carrier phase probe:** Direct SAADC measurement of LF_OA_OUT (AIN5) during emulation confirmed the 125 kHz carrier phase is completely filtered out by the analog peak detector (RC time constant ~2 ms). No PLL or carrier-lock approach is feasible.

**Root cause:** CU's analog frontend has a low-pass filter with ~20 kHz bandwidth, designed for ASK envelope detection. fc/2 = 62.5 kHz is attenuated beyond recovery.

**Solution:** fc/8 (15.625 kHz) passes through the frontend,same frequency range as FSK subcarriers, which work reliably. Switching to fc/8 produced a **300x signal improvement** (IQ magnitude ~89,000 vs ~300). Once the signal was visible, the only remaining issue was searching for the preamble in the differential domain instead of integrating back to raw data via cumulative XOR.

---

### Real Indala reader validation
- **Not tested against a real Indala reader.** The correlation decoder proves the emulated signal carries valid PSK data, but real readers may have different demod sensitivity than PM3. This is the most important gap to close.

### PM3 native PSK demod
- PM3's `lf indala reader` fails with CU's active waveform (~500 errors vs 100 threshold). `lf search` works ~50%.
- If PM3 native support is desired, this likely requires a PM3-side demod improvement; the data IS present, just below PM3's error tolerance. Could be a separate PM3 PR.

### Reader fc/2 limitation
- CU can only **read** Indala at fc/8, not the standard fc/2:
  - CU can read/verify T55XX tags configured for fc/8 (self-test path works)
  - CU **cannot read real Indala tags** (which use fc/2)
  - This is a hardware limitation of the analog frontend (LPF filters out 62.5 kHz)
  - A hardware mod (bypassing the peak detector) could theoretically enable fc/2 reading, but that's out of scope


GameTec-live:

kevihiiin:
> Hey, anyone still working on this?

Hi, sorry life got in the way. I checked it against a physical reader (not PM3), and the tag can be emulated.

The branch has since gone out of sync. I'm happy to resolve the conflicts if you think the approach of the PSK decoder is fine (should be quite robust according to my testing).

GameTec-live:
Ok, cool. Ofc entirely fine, life always first. Ill consider this PR ready again.

imnotverygood:
See https://github.com/RfidResearchGroup/ChameleonUltra/pull/407

LupusE (approved):
I have no access to a physical reader, but approved from code review and dry-test at my CU

LupusE:
Hi @kevihiiin, could you rebase your fork and update the PR? Afterwards we should be able to merge it.

PR #404, opened 2026-04-15
Add Jablotron LF tag support (read, emulate, write to T55xx)

midlan:
After few days with trial and error, I was able to add support for Jablotron LF tags with claude code. To be honest, I am noob at tags cloning/emulation. So review from some skilled person needed. 

Features that I tested:
- tag read
- save tag as Jablotron type to Chameleon slot
- test the tag emulation against real tag reader (door opens)

~~Features that I didn't test:~~
~~- write to T55xx tag~~

EDIT 2026-04-20: today tested write to T5577 tag, it works, jablotron reader reads it
________

Jablotron uses differential biphase (inverted) at RF/64, 64-bit frames:
  bits  0-15: 0xFFFF preamble
  bits 16-55: 40-bit data (5 bytes), bit 16 must be 0
  bits 56-63: 8-bit checksum = (sum of data bytes) XOR 0x3A

Firmware:
  - rfid/nfctag/lf/protocols/jablotron.c  - encoder/decoder codec
  - rfid/nfctag/lf/utils/diphase.c        - inverted-biphase state machine
    (shared util, reusable by other diphase protocols)
  - rfid/reader/lf/lf_jablotron_data.c    - GPIO-interval reader path
  - app_cmd.c: JABLOTRON_SCAN, JABLOTRON_WRITE_TO_T55XX,
                JABLOTRON_SET_EMU_ID, JABLOTRON_GET_EMU_ID
  - tag_base_type.h: TAG_TYPE_JABLOTRON enum
  - t55xx.h: T5577_JABLOTRON_CONFIG (DIPHASE modulation, RF/64)
  - lf_tag_em.c: load callback, factory-default data, save callback

Python CLI (software/script/):
  - lf jablotron read               - scan a real tag
  - lf jablotron write --id         - clone onto T55xx
  - lf jablotron econfig -s N --id  - set emulator ID on a slot
  - hw slot list shows Jablotron ID and decimal card number

Python test (software/script/tests/test_jablotron_modulator.py):
  Pure-Python round-trip validator that reimplements the modulator and
  diphase decoder, expands PWM entries to an edge stream, and confirms
  the decoded data matches the input.  Regression guard for both the
  firmware's double-frame encoding and the single-frame variant.

Notable PWM design choices:
  - Constant-level diphase encoding uses the same PAC pattern: CC=0 for LOW, CC=counter_top+1 for HIGH.  counter_top=31 gives exactly 32 carrier cycles per half-bit at NRF_PWM_CLK_125kHz.
  - The 64-bit frame is encoded twice in the 256-entry PWM buffer with the internal level variable persisting between the two passes. This is required for clean PWM looping: a single 64-bit diphase frame with an odd number of zero bits ends at a level opposite the starting level, leaving no transition at the loop boundary where the reader expects one.  Encoding twice guarantees a continuous diphase stream regardless of the data's zero-count parity.

Reference: Proxmark3 cmdlfjablotron.c


midlan:
ok after few weeks of testing here is my summary:

Jablotron LF tag emulation confirmed working with readers:
JA-122E - tested by me
JA-123E - tested by me
Kyocera card reader - can't find exact model, tested by me
JA-113E - tested by @Meridian85

partially working:
JA-114E - seems working for me, but does not work for @Meridian85

@GameTec-live can you check the code and merge it? Or what I need to do? Add support to GUI?

GameTec-live:
Hey, thanks for testing, Ill take a look once I have time (probably weekend).

GUI support would be appreciated but is no requirement.

midlan:
@GameTec-live I see the branch have merge conflicts, should I fix them?

and should I also add CHANGELOG.md

to speed up the merge/release?

when merged, I will happily add the GUI support

GameTec-live:
If you still want to work on this, please fix the merge conflicts. Your PR should then be merged soon TM as its one of the PRs to still look at in the current featurefreeze

midlan:
Rebased on main; resolved conflicts. Note: JABLOTRON_SCAN and JABLOTRON_WRITE_TO_T55XX are now at 3019/3020 —
  3016 was taken by LF_T55XX_WRITE from PR #413, and I kept SCAN/WRITE adjacent to match the convention of every other LF protocol.

midlan:
added CHANGELOG entry

midlan:
build failing now, but because of an Actions outage: https://www.githubstatus.com/
I will retrigger it when it will be back

midlan:
ping @GameTec-live :) rebased, built, flashed, tested, CI is green, I think it's ready to merge now :)

GameTec-live:
Cool, thanks! Waiting on @LupusE to give the gtg

LupusE (approved):
Fine for me. Even if I could not test in real life.
Should have no impact at pending PR.

PR #406, opened 2026-04-23
hf 14a sniff improvements for nonce collection and crack, fence to catch missing or blocked mfkey binaries

nieldk:
1. check_tools(), mfkey64 added to the list. It now warns at startup if mfkey64 is missing alongside mfkey32v2 and the other tools.
2. _run_mfkey64() rewritten, instead of returning None for any failure, it now returns one of three sentinel strings: _TOOL_MISSING (binary not on disk), _TOOL_BLOCKED (binary exists but OSError/PermissionError/bad exit code, covers AV quarantine, wrong architecture, and permission denied), or _TOOL_NO_KEY (ran cleanly, no key found). A 30-second timeout is also added so a hung/sandboxed process doesn't stall the CLI.
3. _run_mfkey32v2_sniff() added, a separate wrapper for the sniff path that returns the same three sentinels. The original _run_mfkey32v2() is left untouched since it's used by the detection-log Pool.imap() path which has different error-handling requirements.
4. Nonce cracking block updated, the command strings (mfkey64 ... and mfkey32v2 ...) are always printed regardless of whether the binaries can run, so the user can copy and execute them manually.


nieldk:
hf_14a_sniff

software/script/chameleon_cli_main.py
LupusE, line 1: In general i would agree, but this causes:
```
$ LANG=C ./software/script/chameleon_cli_main.py
zsh: permission denied: ./software/script/chameleon_cli_main.py
```

744 would be better?

nieldk: Changed, even had no idea this was possible ….

LupusE (approved):
Checked, works in my tests.
I've done no negative tests (remove the binary manually)

LupusE:
From Discord, maybe one or another BLE issue will be fixed here:

Q:
> Could you elaborate
> - static nrf_saadc_value_t adc_buf[ADC_BUF_SIZE][ADC_BUF_COUNT];
> + static nrf_saadc_value_t adc_buf[ADC_BUF_COUNT][ADC_BUF_SIZE];
> 
> in 406? 
>I haven't take a look into the function itself, is it a bug or an improvement? Could it solve other BLE issues we've seen in the past?

A:
> This is a bug fix. The original declaration adc_buf[ADC_BUF_SIZE][ADC_BUF_COUNT] allocates ADC_BUF_SIZE rows of ADC_BUF_COUNT elements each. Since ADC_BUF_COUNT is 2 (double-buffering), each row is only 2 samples wide. > When the SAADC DMA callback calls nrfx_saadc_buffer_convert(&adc_buf[idx][0], ADC_BUF_SIZE), it passes a pointer to a 2-element slot but tells the hardware to fill ADC_BUF_SIZE samples into it, a silent buffer overflow into adjacent memory on every capture.
> 
> Swapping to adc_buf[ADC_BUF_COUNT][ADC_BUF_SIZE] gives 2 rows of ADC_BUF_SIZE samples each, which is what the double-buffer pattern requires. &adc_buf[idx][0] now correctly points to a full-sized DMA buffer.
> 
> As for BLE, it's plausible. The overflow was scribbling over whatever follows adc_buf in BSS on every LF raw capture, which could corrupt BLE stack state depending on the linker layout. I haven't confirmed that correlation directly, but it would explain intermittent BLE instability observed during LF operations.

PR #367, opened 2026-03-03
feat(lf): add ioProx support (read, emulate, clone)

bernadic:
## Summary
- Adds ioProx LF card support: decode/read, tag emulation, and T55xx cloning.
- Integrates ioProx into the LF reader and tag emulation pipeline.
- Exposes ioProx commands via the Python CLI scripts.

## Testing
- Read/decode tested with real ioProx cards.
- Verified against Proxmark3 emulation.
- Emulation tested on a real door reader and validated with Proxmark3.
- Cloning tested by writing to a T55xx tag and verifying on Proxmark3.

## Notes
- Commits are kept granular and build-safe.

PR #407, opened 2026-04-24
Feature/lf: add IDTECK PSK1 tag emulation and T55xx clone

matteoscrugli:
# lf: add IDTECK PSK1 tag emulation and T55xx clone

## TL;DR

This PR adds **IDTECK** (a 125kHz PSK1 card format) to the supported LF protocols, covering **tag emulation** and **T55xx cloning**. Tested end-to-end against a real IDTECK reader.

## Why this was worth implementing

The Chameleon Ultra already supports a handful of LF protocols (EM410x, HID Prox, ioProx, Viking, PAC), all of them **ASK** or **FSK** modulations. IDTECK belongs to the **PSK1 family**, which on first inspection seemed out of reach: the firmware comment in `lf_sense_enable` noted that the tag-mode antenna taps are envelope-only, which rules out carrier phase-lock and makes coherent demodulation (the approach Proxmark3 uses) unavailable.

The insight that made this PR possible: **PSK1 in practice is a differential encoding**. What the reader decodes are 180° phase transitions *between consecutive bits*, not absolute phase. Other LF tooling (Flipper Zero, for example) chooses to phase-lock anyway for deterministic bit timing, typically by clocking a timer off the reader's carrier via a GPIO, a solution that presupposes a dedicated carrier-sense pin. The Chameleon Ultra's tag-mode front-end does not expose such a tap to the MCU, so that path isn't available; but given that PSK1 is differential, strict phase-lock turns out to be a convenience rather than a requirement. A free-running subcarrier generated from HFXO (±40 ppm) stays well within the PLL tolerance of consumer-grade readers, because the drift between one bit and the next is below 3°.

In other words: we can emit PSK1 from the same `LF_MOD` → load-modulation path the Chameleon already uses for FSK protocols, by simply flipping the PWM polarity at bit transitions, with no carrier sync, no active RF driving, and no hardware modifications.

### Shared PSK1 helper in `utils/psk1.{c,h}`

The subcarrier generator is factored into a reusable helper: given a frame (MSB-first bytes) and a bit count, it fills a `nrf_pwm_values_wave_form_t` buffer with PSK1-encoded entries. The helper lives in `utils/psk1.{c,h}` so future PSK1 protocols (Indala, Keri, NexWatch, ...) can plug in without reimplementing the timing and differential-encoding logic.

### PWM base clock selector

The 16µs subcarrier period (carrier/2 = 62.5 kHz) requires a `counter_top` below 3 at the legacy 125kHz PWM base used by ASK/FSK protocols, which is out of spec for the nRF52 PWM peripheral. Rather than rescaling every existing protocol, `pwm_init` now selects the base clock from the currently-loaded tag type:

```c
cfg.base_clock = IS_PSK1_TYPE(m_tag_type) ? NRF_PWM_CLK_1MHz : NRF_PWM_CLK_125kHz;
```

`IS_PSK1_TYPE()` is a macro in `tag_base_type.h`; today it matches only `TAG_TYPE_IDTECK`, but extends trivially to other PSK1 types in future PRs. **Legacy protocols are untouched, so there is zero regression risk on EM410x, HID Prox, ioProx, Viking, PAC.**

The comment in `lf_sense_enable` that previously implied PSK was out of reach is updated to reflect the actual constraint (no coherent demod) without overstating the limit for differential encodings.

### IDTECK protocol in `protocols/idteck.{c,h}`

The IDTECK frame layout (32-bit fixed preamble `0x4944544B` "IDTK" + 32-bit card payload consisting of a checksum byte and a byte-reversed 24-bit card number) matches what the Proxmark3 client expects, so we emit byte-for-byte what `cmdlfidteck.c` decodes. The `idteck_t55xx_writer` produces the block layout for cloning onto a T55xx tag using `T5577_MODULATION_PSK1 | T5577_BITRATE_RF_32`.

### Command layer and CLI

Three new `DATA_CMD_*` codes, three firmware handlers, and CLI commands:

| Command | Effect |
|---|---|
| `lf idteck econfig -s <slot> --id <hex>` | Set the emulated 64-bit frame on a slot |
| `lf idteck econfig -s <slot>` | Read it back, showing preamble, payload, decoded card number and checksum status |
| `lf idteck write --id <hex>` | Clone the frame to a T55xx tag in reader mode |
| `lf clone -t idteck --id <hex>` | Same write via the unified clone command |
| `hw slot list` | Now renders `Frame:` and `Card ID:` lines for IDTECK slots |

The `--id` argument accepts either 16 hex characters (the full 64-bit frame) or 8 hex (the 32-bit payload, with the preamble auto-prepended for the user's convenience).

The CLI emits a non-blocking informational note when the payload checksum does not match the value computed from the card number, since some readers validate it and some do not. This turns out to matter for a common variant of IDTECK cards in the wild whose checksum byte differs from the simple sum-of-bytes formula.

As with the other per-protocol T55xx writers (`write_em410x_to_t55xx`, `write_hidprox_to_t55xx`, etc.), `write_idteck_to_t55xx` lives in `lf_reader_main.c` and is therefore only compiled on Chameleon Ultra; the corresponding `cmd_processor_idteck_write_to_t55xx` is wrapped in `#if defined(PROJECT_CHAMELEON_ULTRA)` to match the existing convention. The IDTECK emulation path (set/get emu id, slot integration) is available on both Ultra and Lite.

### Not included (intentionally)

- **IDTECK read**: the emulation ADC path samples at 125kHz after an envelope-detection analog front-end, which is insufficient for PSK demodulation via coherent sampling. A dedicated edge-timing decoder on `LF_OA_OUT` is feasible purely in software, reusing the GPIOTE+TIMER edge-capture infrastructure already in place for the Viking reader, not unlike what comparator-plus-input-capture LF tools do. It's left as a follow-up PR to keep review size manageable.
- **Indala / Keri / NexWatch**: the `utils/psk1` helper and `IS_PSK1_TYPE()` macro are designed to absorb these with a one-file-per-protocol pattern, but adding them here would inflate the PR. Follow-up PRs, one per protocol.

## How it was tested

- **Emulation against a consumer IDTECK reader**: the Chameleon was accepted cleanly, reader reacts identically to a real card.
- **T55xx clone**: a blank T5577 fob programmed via `lf idteck write` works on the same reader as expected.
- **Non-regression on other LF protocols**: before switching to IDTECK, other slots already configured with EM410x cards continued to work against their respective readers; cycling via the physical button between an IDTECK slot (1MHz PWM base) and legacy-protocol slots (125kHz PWM base) shows `pwm_init` correctly switches the base clock in both directions, with emulation of the pre-existing cards unchanged.
- **Persistence**: using `hw dfu` + a reflash of the same firmware image as a clean cold-boot proxy (FDS preserved), the slot configuration survives intact, with tag type, emulated frame and nick name all readable via `lf idteck econfig` after reboot.
- **CLI edge cases**: 8-hex and 16-hex inputs both work (with and without the auto-prepended preamble), malformed hex is rejected with a clean error, missing `--id` on `write` is caught by the required-argument validator, and checksum-mismatch frames still go through with a clear warning.
- **Firmware builds clean at every commit** on both targets: the split into 5 commits is bisectable; each step compiles to a working image in isolation on Chameleon Ultra and Lite alike.

## Commit structure

Intentionally split into logical commits (same style as PR #367 for ioProx):

1. `feat(lf): add shared PSK1 wave-form helper for tag emulation`: the reusable subcarrier generator, alone. Compiles but has no caller yet.
2. `feat(lf): add IDTECK tag emulation (PSK1 RF/32)`: the protocol file, the PWM base-clock switch, the tag-emulation handler map wiring, and the T5577 config constant.
3. `feat(lf): integrate IDTECK into firmware command and T55xx write paths`: `DATA_CMD_*`, `app_cmd.c` handlers, `write_idteck_to_t55xx`. After this the firmware is functional end-to-end from the host protocol point of view.
4. `feat(cli): add lf idteck subgroup and extend lf clone with idteck type`: the Python CLI layer and `hw slot list` rendering.
5. `docs(changelog): add IDTECK LF protocol entry`.

Happy to answer any questions during review, especially around the base-clock isolation choice and the decision to defer read support.


LupusE (approved):
Tested against PM3 -> works

Testpattern:
```
[USB] chameleon --> lf idteck econfig -s 4 --id DEAD1234
note: payload checksum 0xDE does not match computed 0xF3 (some readers ignore this, some may reject)
 - IDTECK emu id set to 4944544BDEAD1234.
[USB] chameleon --> hw slot store
 - Store slots config and data from device memory to flash success.
```

Result:
```
[+] IDTECK Tag Found: Card ID 3412653 ( 0x3412AD ) Raw: 4944544BDEAD1234  chksum 0xDE ( fail )
[+] [H10301  ] HID H10301 26-bit                FC: 26  CN: 2390  parity ( fail )

[+] Valid Idteck ID found!
```


Adding `lf idteck read` would be nice for consistency in the LF commnds. But as I am not able to do it, this is just a nice question :)

LupusE:
```
    IDTECK_WRITE_TO_T55XX = 3017
[...]
    IDTECK_SET_EMU_ID = 5010
    IDTECK_GET_EMU_ID = 5011
```

https://github.com/RfidResearchGroup/ChameleonUltra/pull/407/changes#diff-8634035847f2b8ee5f62fd448cc8ef60ad40a6bef432b153d560bef0811db872R96
https://github.com/RfidResearchGroup/ChameleonUltra/pull/407/changes#diff-8634035847f2b8ee5f62fd448cc8ef60ad40a6bef432b153d560bef0811db872R162
https://github.com/RfidResearchGroup/ChameleonUltra/pull/407/changes#diff-8634035847f2b8ee5f62fd448cc8ef60ad40a6bef432b153d560bef0811db872R163

Could you maybe change this part to a higher value? Else it will conflict with PR 404 (https://github.com/RfidResearchGroup/ChameleonUltra/pull/404/)changes#diff-8634035847f2b8ee5f62fd448cc8ef60ad40a6bef432b153d560bef0811db872R153 which was first ...

```
    JABLOTRON_WRITE_TO_T55XX = 3017
[...]
    JABLOTRON_SET_EMU_ID = 5010
    JABLOTRON_GET_EMU_ID = 5011
```

matteoscrugli:
> ```
>     IDTECK_WRITE_TO_T55XX = 3017
> [...]
>     IDTECK_SET_EMU_ID = 5010
>     IDTECK_GET_EMU_ID = 5011
> ```
> 
> https://github.com/RfidResearchGroup/ChameleonUltra/pull/407/changes#diff-8634035847f2b8ee5f62fd448cc8ef60ad40a6bef432b153d560bef0811db872R96 https://github.com/RfidResearchGroup/ChameleonUltra/pull/407/changes#diff-8634035847f2b8ee5f62fd448cc8ef60ad40a6bef432b153d560bef0811db872R162 https://github.com/RfidResearchGroup/ChameleonUltra/pull/407/changes#diff-8634035847f2b8ee5f62fd448cc8ef60ad40a6bef432b153d560bef0811db872R163
> 
> Could you maybe change this part to a higher value? Else it will conflict with PR 404 (#404 which was first ...
> 
> ```
>     JABLOTRON_WRITE_TO_T55XX = 3017
> [...]
>     JABLOTRON_SET_EMU_ID = 5010
>     JABLOTRON_GET_EMU_ID = 5011
> ```

Sure, I'll sort it out tonight or tomorrow, thanks

software/script/chameleon_enum.py
```diff
@@ -158,6 +159,8 @@ class Command(enum.IntEnum):
     PAC_GET_EMU_ID = 5007
     IOPROX_SET_EMU_ID = 5008
     IOPROX_GET_EMU_ID = 5009
+    IDTECK_SET_EMU_ID = 5010
```
LupusE:
```suggestion
    IDTECK_SET_EMU_ID = 5012
```

software/script/chameleon_enum.py
```diff
@@ -158,6 +159,8 @@ class Command(enum.IntEnum):
...
+    IDTECK_GET_EMU_ID = 5011
```
LupusE:
```suggestion
    IDTECK_GET_EMU_ID = 5013
```

software/script/chameleon_enum.py
```diff
@@ -93,6 +93,7 @@ class Command(enum.IntEnum):
     IOPROX_DECODE_RAW = 3012
     IOPROX_COMPOSE_ID = 3013
     LF_T55XX_WRITE = 3016
+    IDTECK_WRITE_TO_T55XX = 3017
```
LupusE:
```suggestion
    IDTECK_WRITE_TO_T55XX = 3018
```

matteoscrugli:
Done! IDs shifted as requested in commit `3b67f50`:

- `IDTECK_WRITE_TO_T55XX`: 3017 → 3018
- `IDTECK_SET_EMU_ID`: 5010 → 5012
- `IDTECK_GET_EMU_ID`: 5011 → 5013

Also rebased on current `main` to pick up the t55write merge (one manual conflict on `LFT55xxClone.TYPES`, resolved by union: `em410x, electra, hid, ioprox, pac, viking, idteck`).

Smoke tested end-to-end on hardware: EM410X non-regression OK, IDTECK PoC against reader OK, T55xx writer OK. The new `FLAG_LOOP` playback coexists cleanly with the dynamic `base_clock` switch in `pwm_init`.

GUI PR GameTec-live/ChameleonUltraGUI#844 updated in cascade and verified live.

For `lf idteck read`: prefer a follow-up PR, the PSK edge-timing demod is non-trivial DSP and would balloon the diff.

LupusE:
2 hours ago we merged an important LF fix into main ... Have you rebased your PR, so #420 won't get overwritten via force-push?
I was just about to review if the fix will conflict with your PR. But both are huge changes, I am afraid I am overlooking something by manual review, so a dialog could close this gap :)

matteoscrugli:
Yes, the force-push is a rebase on top of `013f1f5` (the #420 merge), so the LED fix is preserved. HEAD is `1e78976`.

One manual conflict on `software/script/chameleon_cmd.py` (a duplicated `return resp` after the new `ioprox_get_emu_id` unpack landed adjacent to the IDTECK additions), resolved by union. No scope changes. Smoke tested on hw_v1: slot 3 IDTECK renders and emulates correctly with the new firmware.

P.S. I hadn't seen your message :) I was already working on resolving the issue on my own

LupusE (approved):
Quick retest because of the latest developments. Seems everything fine for me.

jack2game:
Thank you for this PR. Do you have any plan to implement the Indala support?

PR #413, opened 2026-05-01
T55write

nieldk:


GameTec-live:
Whats this?

nieldk:
> Whats this?

Trying to improve some timing in LF

nieldk:
Auth trace, usefull as a tool for having CU (Ultra) act as a reader (for example i used this to test HF14a sniff)

GameTec-live:
Ah, oki.

The description is empty. thats why I was asking

LupusE (approved):
I can't verify the results by now, but the output looks promising.

Works as shown in the screenshot at my testc ards.

nieldk:
> I can't verify the results by now, but the output looks promising.
> 
> Works as shown in the screenshot at my testc ards.

Latest push I added a timeout. So that the reader will actually wait for (default 3s) an amount of time (don’t need to hold a second CU/tag to the reader CU before issuing hf 14a auth command)

adipierro:
Seems like this PR has introduced an issue when emulating LF tag; `RF_FIELD` LED stays on after the field is no longer there; neither the device goes to sleep until the slot is changed.

adipierro:
Unfortunately I don't have a devboard, so can't properly debug this. Will try a different reader soon to understand if the lack of field loss depends on it.

matteoscrugli:
Hi @nieldk, flagging a regression that came in with this PR. The headline T5577 write work is unrelated, but the same change set also rewrote `firmware/application/src/rfid/nfctag/lf/lf_tag_em.c` (commit `f7feda5`, "Add files via upload"). That rewrite broke field-loss detection during LF tag emulation, which is the function that turns the RF status LED off, stops `g_is_tag_emulating`, and arms the sleep timer. Losing it is more than cosmetic: the device never realizes the reader is gone.

## Symptom

After the device enters tag emulation and the reader is then removed, the white RF LED stays on indefinitely, `g_is_tag_emulating` stays true, and the sleep timer never starts. The only way to clear the state is changing slot via the physical button (which calls `lf_sense_disable` explicitly).

## Root cause

The rewrite switched playback from `(LF_125KHZ_BROADCAST_MAX, NRFX_PWM_FLAG_STOP)` to `(1, NRFX_PWM_FLAG_LOOP)`, and moved the field check from `NRFX_PWM_EVT_STOPPED` (with `ANT_NO_MOD()` plus `bsp_delay_ms(1)` settling) to `NRFX_PWM_EVT_END_SEQ0` (no settling).

The previous logic worked because it sampled `LF_RSSI` after explicitly releasing the tank and letting the LC decay. The new logic samples while the PWM is still actively load modulating, so `LF_OA` keeps oscillating from the modulation itself and LPCOMP reads "field present" regardless of whether the reader is still there. `is_lf_field_exists()` returns true forever, `lf_field_lost()` is never reached, and `TAG_FIELD_LED_OFF()` never executes.

## Why the obvious patch (insert `ANT_NO_MOD` plus delay inside `EVT_END_SEQ0`) does not fix it

I tried that first. `NRFX_PWM_FLAG_LOOP` means the PWM peripheral keeps writing `LF_MOD` autonomously. Clearing the pin from the event handler is overwritten by the next PWM cycle a few microseconds later, the tank does not decay, the sample is still dominated by ringing. Confirmed on hardware with 200 us and again with 1 ms delays: LED still stuck.

In other words, the "no gap between sequence plays" optimization that motivated the rewrite is structurally incompatible with the only field-detection mechanism available on this hardware. You cannot sample `LF_RSSI` while load modulation is active. Daniel Wagner's pre-merge logic was not gratuitous: the burst-then-stop-then-sample pattern exists precisely to give the tank a quiet window.

## Why this matters

Beyond the LED, `lf_field_lost()` is the only path that:

* unsets `g_is_tag_emulating` (used elsewhere as a global guard),
* re-arms `LPCOMP INTENSET = LPCOMP_INTENSET_UP_Msk` so the next field arrival can wake the emulator,
* starts `sleep_timer_start(SLEEP_DELAY_MS_FIELD_125KHZ_LOST)` to let the device enter low-power sleep.

With the current code, none of those happen until the user changes slot, so the device stays in an emulation state it can never exit on its own. Battery impact aside, the very ability to detect that the reader has gone away is lost.

## Proposed fix

Partial revert of the relevant chunks of `lf_tag_em.c`. Restore `LF_125KHZ_BROADCAST_MAX (10)`, switch the playback call back to `(LF_125KHZ_BROADCAST_MAX, NRFX_PWM_FLAG_STOP)`, and reinstate the original `EVT_STOPPED` handler with `ANT_NO_MOD()`, `bsp_delay_ms(1)`, LPCOMP int clear, then sample. Same logic that has been running unchanged in the codebase from `a421e99` (March 2026) up to the merge of this PR. Small diff (one file, around 25 net lines), tested on hardware: LED turns off promptly after reader removal, EM410X emulation otherwise unaffected.

Alternatives that would let you keep `FLAG_LOOP` exist (a TIMER plus PPI scheme to insert programmatic quiet windows, or a "silent tail" appended to every protocol's wave form), but they are more invasive than the revert and provide no measurable benefit on the readers I tested. Happy to open them as a separate "perf(lf): gapless emulation" follow-up if there is a use case that motivates it.

If acceptable, I will open a fix PR with the patch and a repro reference back to this one.

Thank you all

@LupusE, @GameTec-live

nieldk:
This is exactly the analysis i arrived at too. the same root cause and same fix i implemented. My lf_tag_em.c already has:	•	NRFX_PWM_FLAG_LOOP → NRFX_PWM_FLAG_STOP with 10 iterations	•	Field check moved back to EVT_STOPPED	•	ANT_NO_MOD() + bsp_delay_ms(2) before samplingThe only difference is i use bsp_delay_ms(2) vs the bsp_delay_ms(1), either is fine given the ~2ms time constant.But feel free to commit a PR :)On 14 May 2026, at 16.31, Matteo Scrugli ***@***.***> wrote:matteoscrugli left a comment (RfidResearchGroup/ChameleonUltra#413)
Hi @nieldk, flagging a regression that came in with this PR. The headline T5577 write work is unrelated, but the same change set also rewrote firmware/application/src/rfid/nfctag/lf/lf_tag_em.c (commit f7feda5, "Add files via upload"). That rewrite broke field-loss detection during LF tag emulation, which is the function that turns the RF status LED off, stops g_is_tag_emulating, and arms the sleep timer. Losing it is more than cosmetic: the device never realizes the reader is gone.
Symptom
After the device enters tag emulation and the reader is then removed, the white RF LED stays on indefinitely, g_is_tag_emulating stays true, and the sleep timer never starts. The only way to clear the state is changing slot via the physical button (which calls lf_sense_disable explicitly).
Root cause
The rewrite switched playback from (LF_125KHZ_BROADCAST_MAX, NRFX_PWM_FLAG_STOP) to (1, NRFX_PWM_FLAG_LOOP), and moved the field check from NRFX_PWM_EVT_STOPPED (with ANT_NO_MOD() plus bsp_delay_ms(1) settling) to NRFX_PWM_EVT_END_SEQ0 (no settling).
The previous logic worked because it sampled LF_RSSI after explicitly releasing the tank and letting the LC decay. The new logic samples while the PWM is still actively load modulating, so LF_OA keeps oscillating from the modulation itself and LPCOMP reads "field present" regardless of whether the reader is still there. is_lf_field_exists() returns true forever, lf_field_lost() is never reached, and TAG_FIELD_LED_OFF() never executes.
Why the obvious patch (insert ANT_NO_MOD plus delay inside EVT_END_SEQ0) does not fix it
I tried that first. NRFX_PWM_FLAG_LOOP means the PWM peripheral keeps writing LF_MOD autonomously. Clearing the pin from the event handler is overwritten by the next PWM cycle a few microseconds later, the tank does not decay, the sample is still dominated by ringing. Confirmed on hardware with 200 us and again with 1 ms delays: LED still stuck.
In other words, the "no gap between sequence plays" optimization that motivated the rewrite is structurally incompatible with the only field-detection mechanism available on this hardware. You cannot sample LF_RSSI while load modulation is active. Daniel Wagner's pre-merge logic was not gratuitous: the burst-then-stop-then-sample pattern exists precisely to give the tank a quiet window.
Why this matters
Beyond the LED, lf_field_lost() is the only path that:

unsets g_is_tag_emulating (used elsewhere as a global guard),
re-arms LPCOMP INTENSET = LPCOMP_INTENSET_UP_Msk so the next field arrival can wake the emulator,
starts sleep_timer_start(SLEEP_DELAY_MS_FIELD_125KHZ_LOST) to let the device enter low-power sleep.

With the current code, none of those happen until the user changes slot, so the device stays in an emulation state it can never exit on its own. Battery impact aside, the very ability to detect that the reader has gone away is lost.
Proposed fix
Partial revert of the relevant chunks of lf_tag_em.c. Restore LF_125KHZ_BROADCAST_MAX (10), switch the playback call back to (LF_125KHZ_BROADCAST_MAX, NRFX_PWM_FLAG_STOP), and reinstate the original EVT_STOPPED handler with ANT_NO_MOD(), bsp_delay_ms(1), LPCOMP int clear, then sample. Same logic that has been running unchanged in the codebase from a421e99 (March 2026) up to the merge of this PR. Small diff (one file, around 25 net lines), tested on hardware: LED turns off promptly after reader removal, EM410X emulation otherwise unaffected.
Alternatives that would let you keep FLAG_LOOP exist (a TIMER plus PPI scheme to insert programmatic quiet windows, or a "silent tail" appended to every protocol's wave form), but they are more invasive than the revert and provide no measurable benefit on the readers I tested. Happy to open them as a separate "perf(lf): gapless emulation" follow-up if there is a use case that motivates it.
If acceptable, I will open a fix PR with the patch and a repro reference back to this one.
Thank you all
@LupusE, @GameTec-live

matteoscrugli:
Thanks @nieldk for the quick confirmation. Glad we converged on the same diagnosis and fix independently. Took your suggestion on the delay value, the PR now uses `bsp_delay_ms(2)` to align with the ~2 ms RC time constant of the LF_RSSI peak detector. 1 ms also tested working on hardware, but 2 ms gives a more comfortable margin at negligible cost.

PR #417, opened 2026-05-06
fix: hf mf eview param error due to chunk exceeding 32-block limit

### LINKED ISSUE #412: hf mf eview not working
The command hf mf eview seems to be not working both with default parameter (slot omitted) and slot declared.

Image

azuwis:
The firmware limits mf1_read_emu_block_data to at most 32 blocks per request, but eview's chunk_count only honored data_max_length (256). Added the same 32-block cap already used by esave.

Fix #412


GameTec-live (approved):
sure, lgtm. Still needs a test.

LupusE (approved):
It's the same in line 3756. The values won't be alternated, just the min value of the size is affected ...
Looks fine for me.

PR #418, opened 2026-05-06
Align hf 14a raw CLI parameters with proxmark3

azuwis:
Make `hf 14a raw` take the same flags and positional args as pm3.

Changes:

- `-d <hex>` → positional `<hex> [<hex>...]`
  Before: `hf 14a raw -b 7 -d 40 -k`
  After:  `hf 14a raw -b 7 40 -k`

- Added `-3` / `--no-rats` — 14443-3 select, skip RATS
- Added `-v` / `--verbose`
- Firmware side gets a `no_rats` bit in the payload, temporarily switches
  `forcerats` config during card select

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project uses the changelog in accordance with [keepchangelog](http://keepachangelog.com/). Please use this to write notable changes, which is not the same as git commit log...
 
 ## [unreleased][unreleased]
+ - Changed `hf 14a raw` CLI parameters to align with Proxmark3: replaced `-d`/`--data` with positional hex args, added `-3`/`--no-rats`, `-v`/`--verbose` (@azuwis)
+ - Added `no_rats` parameter to `hf 14a raw` for ISO14443-3 select only (skip RATS) (@azuwis)
  - Added PAC/Stanley LF protocol support: read, emulate and T55xx clone (@kevihiiin, @danieltwagner)
  - Fix firmware application USB serial number (@taichunmin)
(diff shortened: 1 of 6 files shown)


GameTec-live:
Stupid question: Why?

Whats the benefit of rearranging the parameters and what does norats do?

azuwis:
Two things:

1. Align CLI syntax with pm3's `hf 14a raw` — positional hex args instead of `-d <hex>`, so muscle memory carries over between tools.

2. Add `-3`/`--no-rats` — skips RATS during select, matching pm3's [`-3`][1] flag. Firmware side just temporarily toggles `forcerats = 2` for the select, then restores it.

Happy to adjust if anything looks off or not wanted.

[1]: https://github.com/RfidResearchGroup/proxmark3/blob/a804c64d84b96fa40555eb29010498cc36d48470/armsrc/iso14443a.c#L3507

GameTec-live:
Considering the CU is an entirely different tool, Idk if consistency with the pm3 CLI is such a desired thing. The more important thing would be consistency within the CU CLI and tooling. Idk how it fits in there, so thats something to check.

For the RATs skip, ig thats fine and was a missing option. Ill have to check tho that it doesnt break any clients cause changing an existing API firmware side is risky.

azuwis:
I've made other PR #419 that only keep the bug fix commit d2c1f43a0e21bbfaa5de4c175e0c2f61c2f68ebb

As for the `--no-rats`, we already have `hf 14a config --rats skip` when skipping RATS is needed.

Feel free to close this PR.

GameTec-live:
Oki, prob makes more sense to keep it as is. Ill merge the bugfix instead.

PR #419, opened 2026-05-07
fix: hf14a_raw should return data bytes, not Response object

azuwis:
Callers treat the return value as bytes (len(), slicing), but hf14a_raw was returning the Response object itself, causing TypeError.

Before:

```
[USB] chameleon --> hf 14a raw -sc -d 3000
CLI exception: Traceback (most recent call last):
  File ".../chameleon_cli_main.py", line 146, in exec_cmd
    raise error
  File ".../chameleon_cli_main.py", line 141, in exec_cmd
    unit.on_exec(args_parse_result)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File ".../chameleon_cli_unit.py", line 7143, in on_exec
    if len(resp) > 0:
       ~~~^^^^^^
TypeError: object of type 'Response' has no len()
```

After:

```
[USB] chameleon --> hf 14a raw -sc -d 3000
 - 04
```


GameTec-live (approved):
lgtm

PR #424, opened 2026-05-14
fix: load_key_file and load_dic_file for hf mf fchk

### LINKED ISSUE #409: "hf mf fchk" seems to be patially broken
I run in the CLI, latest version, compiled today from the REPO

`hf mf fchk --1k <key1> <key2> <key3> <key4> `

And get the green table with the checked keys.
Then I run

`hf mf fchk --1k <key1> <key2> <key3> <key4> --export-dic test.dic`

the file is created and contains the same keys
Finally i run 

`hf mf fchk --1k --key test.dic`

And I get an empty red table.

Image
Why are 0 keys checked in fast 10 seconds?

I tried for check

`hf mf fchk --1k --dic test.dic`

and the command gives a <CR> response (no error, no data)

So I suppose the check on keys on file is not working.

azuwis:
load_key_file (`--key`): read raw 6-byte binary keys instead of trying UTF-8 decode, which failed on binary key files.

load_dic_file (`--dic`): implement stub that was just returning keys unchanged; now reads 12-char hex keys line by line.

Fix #409

### DIFF
--- a/software/script/chameleon_cli_unit.py
+++ b/software/script/chameleon_cli_unit.py
@@ -72,17 +72,28 @@ default_cwd = Path.cwd() / Path(__file__).with_name("bin")
 
 def load_key_file(import_key, keys):
     """
-    Load key file and append its content to the provided set of keys.
-    Each key is expected to be on a new line in the file.
+    Load binary key file and append its content to the provided set of keys.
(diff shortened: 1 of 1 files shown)


LupusE (approved):
Needs clarification in the docs. But I understand the change.

PR #425, opened 2026-05-14
fix: fill both key A and key B in sector trailer when dumping

### LINKED ISSUE #410: CLI dump command don't save keys
I try to dump via CLI a mifare with a key file containing the 2 correct keys but keys are not written on the dump. I get "00 00 00 00 00 00 78 77 88 FF 00 00 00 00 00 00" on the 3rd block... (KEY A and B are set to "0")

`chameleon --> hf mf dump -t bin -f dump.bin -d my.dic`

This causes the slot where I upload the file not working.
(The read and save process via MTools BLE saves the keys on the correct place)

If i download a working dump from a slot with

`hf mf esave -s 1 -t bin -f export_dump.bin`

I get the keys, too. I can compare _export_dump.bin_ with _dump.bin_ with

`hf mf view -d export_dump.bin`

And this demonstrate that `hf mf dump` is bug affected.

azuwis:
MIFARE Classic blanks the authenticated key bytes when reading a sector trailer. Find both key A and key B for each sector, then fill both into the trailer from known keys instead of relying on the read response.

Fix #410

### DIFF
--- a/software/script/chameleon_cli_unit.py
+++ b/software/script/chameleon_cli_unit.py
@@ -2992,34 +2992,50 @@ class HFMFDump(MF1AuthArgsUnit):
 
         # iterate over sectors
         for s in range(16):
-            # try all keys for this sector
-            typ = None
+            # find working keys for this sector (both A and B)
+            type_a = type_b = None
+            key_a = key_b = None
             for key in keys:
-                # first try key B
-                try:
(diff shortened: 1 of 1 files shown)


LupusE (approved):
Nice work. 
As it just changed displaying the result, no test at the CU itself.

PR #426, opened 2026-05-14
fix: fall back to hardnested in autopwn when nested fails

azuwis:
When nt_level is 1 (PRNG_WEAK), nested recovery can fail because recover_a_key re-detects the PRNG and returns None if it's actually HardNested (nt_level == 2). Also, the PRNG detection can misclassify a HardNested tag as PRNG_WEAK from a single NT sample.

Now autopwn tries hardnested as a fallback when nested returns None.


LupusE (approved):
I admit, I am not as deep in the attacks as I wish to be. But at least all my test cards are still working at autopwn.

PR #429, opened 2026-05-15
feat: split linter and formatting check

GameTec-live:
Split type linter and formatting check.

Formatting should be mandatory, types right now arent feasible to keep enforced, therefore that job can fail without consequence.


LupusE (approved):
Cut my code into pieces, this is my last debug
 -- Papa Roach

LupusE (approved):
As this is for the git-handling and not the code itself, I have a limited understanding. But the change looks reasonable for better code quality.

PR #431, opened 2026-05-18
Add DESFIRE_AUTH_CHECK, add BL/Free mem info

nieldk:
The existing HF14A_4_READER_APDU command requires the host to make 4 separate round trips per DESFire key attempt over BLE/USB: field-cycle scan, SelectApplication, Authenticate step 1, and Authenticate step 2. At ~30–50 ms BLE latency per trip, a dictionary of 50 AES keys across 5 AIDs takes 30–50 seconds in BLE latency alone, before any NFC time is counted.

This PR introduces DATA_CMD_HF14A_4_DESFIRE_AUTH_CHECK (6006), which performs the complete DESFire mutual authentication sequence inside the firmware in one host round trip, and updates the Python CLI accordingly.

**Firmware changes**
data_cmd.h

DATA_CMD_HF14A_4_DESFIRE_AUTH_CHECK = 6006

app_cmd.c
New command handler cmd_processor_hf14a_4_desfire_auth_check.
Request payload: [algo(1)][key_slot(1)][flags(1)][aid(3)][key(8|16|24)]

algo: 0=DES, 1=2TDEA, 2=AES-128, 3=3K3DES
flags bit 0: skip_scan — skip field cycle + RATS when card is already active in T=CL (safe for consecutive attempts on the same AID, saves ~20–40 ms per attempt)

Response: STATUS_HF_TAG_OK + [0x01] authenticated / [0x00] wrong key; STATUS_HF_TAG_NO if no card.
The handler includes self-contained crypto:

AES-128 encrypt: nRF52840 hardware ECB via sd_ecb_block_encrypt() (SoftDevice-safe)
AES-128 decrypt: software inverse cipher (InvSubBytes / InvShiftRows / InvMixColumns / AddRoundKey); needed because the ECB peripheral is encrypt-only
DES / 2TDEA / 3K3DES: compact software implementation (standard FIPS permutation tables + S-boxes); uses sd_rand_application_vector_get() for RndA generation

The full auth sequence per call: optional field cycle → scan_auto (REQA + anti-collision + RATS) → SelectApplication → Authenticate step 1 → Authenticate step 2 → verify RndA.
Reuses the existing static tcl_apdu_() helper (already in app_cmd.c) for ISO14443-4 I-block framing and chaining.
Registered with before_hf_reader_run / NULL after-hook (field intentionally left on between calls to support skip_scan).
ble_main.c

On BLE_GAP_EVT_DISCONNECTED, if the device is in reader mode, pcd_14a_reader_antenna_off() is now called immediately. Previously the HF antenna could remain powered indefinitely if the host disconnected mid-scan. Guarded by #if defined(PROJECT_CHAMELEON_ULTRA).

**Python CLI changes**
chameleon_enum.py

HF14A_4_DESFIRE_AUTH_CHECK = 6006
HF14A_SET_FIELD_ON = 2100, HF14A_SET_FIELD_OFF = 2101 (existed in firmware dispatch table but were missing from the Python enum)

chameleon_cmd.py

hf14a_4_desfire_auth_check(algo, key_no, aid, key, skip_scan=False), sends the new command, validates parameters, 5 s timeout
hf14a_set_field_off() / hf14a_set_field_on(), thin wrappers for the existing field control commands

chameleon_cli_unit.py
hf des chk:

Replaces the multi-APDU _try_key() (which called _desfire_select_app + one of _desfire_auth_des/aes/3k3des) with a single hf14a_4_desfire_auth_check call. No longer requires the cryptography Python library for key checking.
Tracks skip_scan per AID: first attempt uses False (full activation), subsequent attempts use True (card stays in T=CL).
Main loop wrapped in try/KeyboardInterrupt/finally; hf14a_set_field_off() called in finally so the RF field is killed even on Ctrl+C.

hf des auth (new command):

Direct single-key authentication test: hf des auth --aid 3F6D08 -n 1 -k <hex>
Algo inferred from key length (8 B → DES, 16 B → AES-128, 24 B → 3K3DES) or forced with --type
Calls hf14a_4_desfire_auth_check directly; prints pass/fail in one round trip
Also calls hf14a_set_field_off() in finally.

Tested against MIFARE DESFire EV2 with default AES-128 key (00…00) on AID 000000. Key scan with 50-key AES dictionary over BLE: ~12 s vs ~50 s with the old APDU-passthrough approach.


nieldk:
Added CLI commands, to show Bootloader version, and Free mem

[USB] chameleon --> hw blver
 - Bootloader Version: v0.3
[USB] chameleon --> hw freemem
 - Heap free  : 123,848 bytes
 - Heap used  : 96 bytes
 - Heap total : 123,944 bytes  (0.1% used)

LupusE (approved):
My physical DESFire card will be delivered in August, I can't wait to check this in real life.
But this change works emulated.

PR #432, opened 2026-05-24
feat(standalone): add standalone modes

nieldk:
Adds a pluggable framework for running workflows on the device without a
connected host. Configure once over USB/BLE, arm with a both-button chord,
capture/operate using physical buttons, retrieve results when reconnected.

- **authtrace**, multi-session HF14A reader-side auth-trace capture. Thin
  wrapper over the existing CMD 2017 machinery; `cmd_processor_hf14a_auth_trace`
  is refactored to expose `hf14a_auth_trace_run()` as a reusable function.
- **slot_cycle**, timed emulation-slot rotation. Reference implementation
  for the `standalone_mode_iface_t` contract; no tag interaction.

- **Chord vocabulary only.** Single-button A/B short/long is already owned by
  `SettingsButtonFunction` (CycleSlot, CloneIcUid, etc.).
  Standalone uses both-short (primary), both-long (arm/disarm toggle),
  both-vlong (destructive).

- **Pre-existing button bug fixed as a side effect.** The GPIOTE handler
  shares a single debounce timer between buttons A and B; when both change
  state within the ~50ms debounce window, only the second event's flag gets
  updated. Latent bug, no upstream feature exercised simultaneous-press.
  Fixed in `timer_button_event_handle` via a rescue path that samples the
  other pin's level after the per-pin code. Single-button behavior
  unchanged. .

- **FDS file ID `0x1010`**, state at key `0x0001`, per-mode config at
  `0x0100 + mode_id`. No collision with existing IDs.

- **Mode IDs are persisted** — `STANDALONE_MODE_*` enum values must be
  appended, not renumbered.

- **CMD IDs 7000–7006** for the seven host commands. Clean range above
  the existing 6005.

- **`STANDALONE_FLAG_HOST_OPTED_IN`** gates modes with destructive
  capabilities. Reserved for future autoclone/read_replay; neither shipped
  mode needs it.

Verified on Ultra v2.1.0:
- chord-based arm/disarm and capture
- device-button session capture (not just CLI trigger)
- FDS persistence across reboots
  authtrace sessions against a MIFARE Classic 1K

Three reserved mode IDs (`autoclone`, `read_replay`, `dict_check`) are
declared but not implemented.


nieldk:
Superseded by #445

PR #433, opened 2026-05-24
Hf sniff renamed to hf trace

nieldk:
Hf sniff renamed to hf trace.
Added option for saving/loading captured data


LupusE:
I followed the discussion at discord. Legit change to me, to make more clear what the feature does. Similar to PM3 functions.

LupusE (approved):
See comment before

Issue #334, opened 2026-01-20
Fixed the issue where the battery level could not be displayed as 100% when fully charged.


GH-H4d35:
https://github.com/RfidResearchGroup/ChameleonUltra/blob/7065011f228cbe54b7ce794697067bdec8e7f6b3/firmware/application/src/ble_main.c#L231-L247

I believe the value of #define P100VOL 4200 is incorrect. This is the theoretical upper limit for 90mAh, but the actual value is only between 3160-3190, which will only cause the software to fail to display 100% after the battery is fully charged and then disconnected.

I recommend using the minimum compatible value, and also suggest adding a battery calibration function and a battery health display to help us understand the current state of the battery.

PR #438, opened 2026-06-06
Improve battery calibration and health reporting

alexl83:
## Summary
This PR is inspired by and aims to address [issue #334](https://github.com/RfidResearchGroup/ChameleonUltra/issues/334), improving battery reporting across firmware, CLI, and the on-device UI.

### Battery calibration and percentage mapping
- Calibrates the battery curve against the conservative lower bound of the settled full-charge range so fully charged devices can reach 100% again.
- Clarifies that the previously mentioned 3160 mV value is inconsistent with a fully charged single-cell LiPo and should be treated as a typo unless board-specific evidence says otherwise.

### Battery health hint
- Adds a shared battery-health helper in firmware so the same heuristic is used consistently by the application, the protocol command, and the on-device UI.
- Extends `hw battery` with a backward-compatible battery-info response that includes a qualitative health hint: `excellent`, `good`, `fair`, `low`, or `critical`.
- Updates the on-device battery screen to use health-aware colors and blink cues for `low` and `critical` states.
- Keeps the old battery-info command working for older firmware clients through a graceful fallback.

## Related documentation
- The companion documentation work is tracked separately in [ChameleonUltraDocs PR #8](https://github.com/RfidResearchGroup/ChameleonUltraDocs/pull/8).

## Validation
- Built the firmware successfully with the official `firmware/build.sh` script.
- Regenerated the flashable DFU packages for the Ultra target.
- Verified the Python CLI modules compile successfully with `python3 -m py_compile`.


GameTec-live:
Ok, so this is 100% AI, right? You didnt even read the Readme in the docs folder.

alexl83:
Hey GameTec-live,
Yes it is, codex specifically

I've read the wiki documentation, starting with development as I wanted to compile fw on my laptop.

Once read bug #334 I got hooked as I never got my battery to show 100% charge; up to today I was convinced my kickstarter ChameleonUltra battery was faulty

Nevertheless, I did not read coding guidelines, my fault for not finding/searching thoroughly 

As my only intent is to contribute, should this PR be not acceptable I humbly apologize; don't mean to waste your/your team's time/resources, really. Let me know and I will close this PR.

Please accept my apologies :)

GameTec-live:
Well, I am not against AI code, I often use it myself, but you need to understand it, know what its doing and be responsible for it.

You (or rather the AI) blindly dropping the md file in the docs folder when the readme in there states otherwise and then committing that without moving that shows, that you probably didnt read the output.

I appreciate you wanting to help, but please do so in a responsible manner.

Take a look at your PR, understand the code, clean it up a bit and maybe make sure the files are where they should be.

On functionality: You improving the percentage curve is nice, thx, will be looked at, but why add another function for the whole "low, critical, ..."? The client already recieves the percentage and voltage, so if the client wants to, it is free to display that in whatever way it wants (eg CUGUI with the icons, the CLI using text, ...). What would be the benefit of computing that again and doing so in firmware?

Once youre done, please convert this back from a draft as contributions are always welcome.

alexl83:
Thanks @GameTec-live for your openness, appreciate it! I did not catch the documentation topic, you're correct: moving docs to dedicated repo fork and presenting a cross-referenced PR.

Regarding the 'on-device battery health' functionality, here is the rationale behind it: 

Just had a faulty Chameleon Ultra for the past two weeks, that I just returned: it was working right, but the battery was clearly faulty as it didn't hold a charge for more than a couple days (tops) and would take many hours charging with 'funky' reporting via client/app. 

I'm a heavy user of the 'charge' functionality embedded in fw, mapped it to a button A long press: it took me a while realizing the battery was faulty cause my spot usage of the device, in the past couple weeks always reported plausible charge levels via leds (btw this PR makes led charge levels more precise).

I thought an on-device 'health' reporting, both calculation and display functionality was useful, especially for limited interface/ui present on such an embedded device, and its highly likely headless operations.

Battery health checks are present in many portable devices I own, but the challenge is that our CU can be used for long periods of time completely headless and without connecting to phone/app, unlike an iPhone e.g. that cannot be 'run' without iOS (and thus its battery health reporting system) - this reasoning led me to the double impleemntation on-device and in-protocol for client

Would you consider keeping something as such on-device?

Perhaps the client/protocol implementation is redundant or not so useful from a best practice/design point of view - I addressed both out of overengineering perhaps (and lack of skills)

PR #439, opened 2026-06-07
Improve battery calibration and health reporting

alexl83:
## Summary
This PR is inspired by and aims to address [issue #334](https://github.com/RfidResearchGroup/ChameleonUltra/issues/334), improving battery reporting across firmware, CLI, and the on-device UI.

### Battery calibration and percentage mapping
- Calibrates the battery curve against the conservative lower bound of the settled full-charge range so fully charged devices can reach 100% again.

### Battery health hint
- Adds a shared battery-health helper in firmware so the same heuristic is used consistently by the application, the protocol command, and the on-device UI.
- Extends `hw battery` with a backward-compatible battery-info response that includes a qualitative health hint: `excellent`, `good`, `fair`, `low`, or `critical`.
- Updates the on-device battery screen to use health-aware colors and blink cues for `low` and `critical` states.
- Keeps the old battery-info command working for older firmware clients through a graceful fallback.

## Related documentation
- The user-facing calibration guide is maintained separately in [ChameleonUltraDocs PR #8](https://github.com/RfidResearchGroup/ChameleonUltraDocs/pull/8).
- This firmware PR intentionally stays focused on the battery behavior, protocol, CLI, and on-device UI changes.

## Validation
- Built the firmware successfully with the official `firmware/build.sh` script.
- Regenerated the flashable DFU packages for the Ultra target.
- Verified the Python CLI modules compile successfully with `python3 -m py_compile`.


alexl83:
Thanks @GameTec-live for your openness, appreciate it! I did not catch the documentation topic, you're correct: moving docs to dedicated repo fork and presenting a cross-referenced PR.

Regarding the 'on-device battery health' functionality, here is the rationale behind it:

Had a faulty Chameleon Ultra for the past two weeks, that I just returned: it was working right, but the battery was clearly faulty as it didn't hold a charge for more than a couple days (tops) and would take many hours charging with 'funky' reporting via client/app.

I'm a heavy user of the 'charge' functionality embedded in fw, mapped it to a button A long press: it took me a while realizing the battery was faulty due to my spotty usage of the device, in the past couple weeks always reported plausible charge levels via leds (btw this PR makes led charge levels more precise).

I thought an on-device 'health' reporting, both calculation and display functionality was useful, especially for limited interface/ui present on such an embedded device, and its highly likely headless operations.

Battery health checks are present in many portable devices I own, but the challenge is that our CU can be used for long periods of time completely headless and without connecting to phone/app, unlike an iPhone e.g. that cannot be 'run' without iOS (and thus its battery health reporting system) - this reasoning led me to the double implementation on-device and in-protocol for client

Would you consider keeping something as such on-device?

Perhaps the client/protocol implementation is redundant or not so useful from a best practice/design point of view - I addressed both out of overengineering perhaps (and lack of skills)

LupusE (approved):
Looks clean to me.
Battery tests are not easy as the battery life of the CU is incredible. But at least above 80% it works as described.

PR #441, opened 2026-06-10
fix: static and staticnested return well-formatted keys

triplesprawl:
# Summary
Staticnested and nested return keys in the wrong format (keys are too short) if the keys contain leading zeros.

# Fix
Replace in printf format "%" to "%012"


LupusE (approved):
I don't know what to say. just approved.

PR #442, opened 2026-06-13
feat: MIFARE Ultralight C authentication and emulation

Foxushka:


lcycon:
Works well for me on a real card, though the dump doesn't include blocks 44-48 (the key), which seemed to lead to issues. Once adding in the key to the end of the dump and loading into to the emulator, all is well.

PR #443, opened 2026-06-13
Export low-level ISO14443-4 handling

aaronjamt:
This PR separates the low-level part of the ISO14443-4 simulation code so it can be used for other tags. This means that tags based on ISO14443-4 can be added more easily in the future, only having to handle the (r)APDUs themselves.

Note that this should not change any existing behavior.


LupusE (approved):
No breaks of anything existing in my tests.

PR #446, opened 2026-06-16
Fix ISO14443-4 block values

aaronjamt:
Was having issues with getting WTX working and realized `PCB_SBLOCK_WTX` was incorrect. The ISO14443-4 standard counts from 1, so a byte looks like `87654321`. Looks like this caused confusion and caused `PCB_CID_FOLLOWING`, `PCB_NAD_FOLLOWING`, and `PCB_CHAIN` to be shifted over by one bit. I also realized `PCB_RBLOCK_MASK`/`PCB_RBLOCK_VAL` were wrong, but unused.


LupusE (approved):
Hard to verify standalone, makes sense with #443 . Good to go.

PR #448, opened 2026-06-17
hf 14a sniff: add parity checks

DidierA:
Adds exclamation marks in the output of `hf 14a sniff`, on bytes where  the parity check fails .

### DIFF
--- a/software/script/chameleon_cli_unit.py
+++ b/software/script/chameleon_cli_unit.py
@@ -31,6 +31,7 @@ from chameleon_utils import (
     execute_tool,
     tqdm_if_exists,
     print_key_table,
+    odd_parity_byte,
 )
 from chameleon_utils import CLITree
 from chameleon_utils import CR, CG, CB, CC, CY, C0, color_string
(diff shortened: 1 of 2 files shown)


LupusE (approved):
Changes only in the CLI, seems reasonable.

software/script/chameleon_cli_unit.py
```diff
@@ -7843,8 +7848,11 @@ def on_exec(self, args: argparse.Namespace):
         last_auth_keytype = None
         last_auth_block = None
 
-        for n, (szBits, data, is_tx) in enumerate(frames):
-            hex_str = ' '.join(f'{b:02x}' for b in data)
+        for n, (szBits, data, is_tx, parity_bits) in enumerate(frames):
+            if (len(data) == len(parity_bits)):
+                hex_str = ' '.join(f"{b:02x}{'!' if odd_parity_byte(b)!= p else ''}" for (b,p) in zip(data,parity_bits))
```
unkernet: I suggest to add space placeholder if parity bit is correct, like proxmark does:
```c
hex_str = ' '.join(f"{b:02x}{'!' if odd_parity_byte(b)!= p else ' '}" for (b,p) in zip(data,parity_bits))
```

Also I suggest to calculate parity bits if it is missing for the specified frame. This will allow to use same print routine and will make future trace export to the PM3 format easier.

LupusE:
Hi @DidierA, could you implement the suggestion from @unkernet ? I've tested it and it works at my system.
Than the PR needs some love (merge conflicts), because we implemented other PRs in between. I don't have the permission to see what exactly is the conflict, and I think you could easily rebase your fork, which is faster than me cloning the whole project to analyze.

I really want this PR merged, as it makes the UI even better.

software/script/chameleon_cli_unit.py
DidierA: I have added the space for alignment. As for adding the missing parity bits, `hf sniff` should display what really happened, not what maybe should have been sent, so if we're going to print the missing bits, it should be toggled by an option, and clearly stated on the display.

Issue #444, opened 2026-06-14
hf 14a sniff bug?


DidierA:
**Describe the bug**
The sniff and dump feature seems to misinterpret some of the data, and save wrong values. Using proxmark3 to read a sector from a mifare 1k slot, and comparing the traces give different values.

**To Reproduce**

1. Import a mifare 1k card to slot 1, with default keys
2. Place the chameleon ultra on the proxmark3
3. In hf cli, type `hf 14a sniff`
4. In pm3, type `hf mf rdsc --sec 1`
5. In pm3, type `trace list -c -t mf`
6. Note the differences, in the card responses

**Expected behavior**
I should observe the same values in the proxmark 3 trace and Chameleon trace (being aware that proxmark's app is able to decode the encrypted part of mifare transmission, the differences can be seen in the encrypted part (ie even using `trace list -t 14a` to have no interpretation of the mifare commands)

**Screenshots**
on pm3:
```
      Start |        End | Src | Data (! denotes parity error)                                           | CRC | Annotation
------------+------------+-----+-------------------------------------------------------------------------+-----+--------------------
          0 |        992 | Rdr |52(7)                                                                    |     | WUPA
       2116 |       4484 | Tag |04  00                                                                   |     | 
       7040 |       9504 | Rdr |93  20                                                                   |     | ANTICOLL
      10548 |      16436 | Tag |E3  B7  55  9F  9E                                                       |     | 
      19584 |      30048 | Rdr |93  70  E3  B7  55  9F  9E  37  84                                       |  ok | SELECT_UID
      31172 |      34692 | Tag |08  B6  DD                                                               |  ok | 
      37504 |      42272 | Rdr |60  04  D1  3D                                                           |  ok | AUTH-A(4)
      44084 |      48756 | Tag |9E  C5  66  9A                                                           |     | AUTH: nt (lfsr16 index 60005)
      58240 |      67616 | Rdr |A7! 27! 02  05! F4! 8F  1E! 52                                           |     | AUTH: nr ar (enc)
      70452 |      75124 | Tag |EB! ED  F5  2B                                                           |     | AUTH: at (enc)
      81792 |      86560 | Rdr |11! 2B! 32! 65!                                                          |     | 
            |            |  *  |                                              key FFFFFFFFFFFF prng WEAK |     |
            |            |  *  |30  04  26  EE                                                           |  ok | READBLOCK(4)
[...]

PR #445, opened 2026-06-15
Standalone BLE relay mode + DESFire AES auth-trace + crypto foundation

nieldk:
Adds a transparent two-device ISO 14443-4 relay over connectionless BLE, plus the DESFire tooling needed to exercise and verify it. These land together because they’re interdependent: the relay is validated via hf des auth-trace, which depends on the DESFire crypto layer, and all three share changes in the NFCT layer (nfc_14a.c).

Legacy 31-byte advertising truncated DESFire frames; extended advertising raises the payload to 245 bytes.

ISO 14443-4 WTX to absorb BLE round-trip latency within the reader’s frame-waiting time.

UID-only relay (4/7-byte), RATS/ATS, single-command 14443-4 exchanges

Round 1 of DESFire AES auth end to end (full 91AF + E(RndB))

Multi-round DESFire AES auth is bounded by the real card’s fixed FWT (~77ms on the test EV3); the BLE round-trip between rounds can exceed it. Single-command ops are unaffected.

MIFARE Classic crypto relay out of scope (CRYPTO1 FWT ≪ BLE latency); UID-only MFC works.

PR #449, opened 2026-06-18
nfc_tag_14a_event_callback(): Fixes wrong reported length when sniffing

DidierA:
In `nfc_tag_14a_event_callback()`, when sniffing, a computation was made to tell which number of parity bits is in the array, and provide the number of bits to` m_tx_sniff_cb()`. However this is not necessary, since `amt` already contains the number of bits, and the computation is wrong: it extracts the 3 least significant bits of `amt` to get how many parity bits there are, but there can be more than 7 parity bits.

This fixes issue #444 .


nieldk:
For the reviewers mostly.

Just a comment , my pr #445 standalone-relay independently reworked this sniff path with the szFrameBits/9 approach.

Conflicts might arise.

please merge this BEFORE #445!
Then I will align

DidierA:
Hi,
Anything else required for this PR to be merged? Let me know if I can help.

PR #451, opened 2026-06-24
perf(lf): drop double-precision sqrt from FSK demod hot path

wereii:
Found this performance tweak while working with Claude on adding the Pyramid tag - tested on my Chameleon Ultra.
Not required for anything else.

---
Claude:

The build targets the Cortex-M4F single-precision FPU (-mfpu=fpv4-sp-d16), which has no double-precision hardware. goertzel_mag() used sqrt() (double): the float result was promoted to double for the call and converted back, and a software double-sqrt routine ran -- twice per decoded bit.

The bit decision only compares the two Goertzel outputs, and power is monotonic with magnitude, so the sqrt is unnecessary. goertzel_power() returns the squared magnitude and fsk_feed() compares that directly. This also removes a latent sqrt(NaN): the magnitude argument can round slightly negative near zero signal, which sqrt() turned into NaN (and NaN comparisons make the bit decision unreliable); comparing the raw power is well-defined.

The old goertzel_mag() had no callers anywhere and was not declared in the header, so it is removed rather than kept.

Shared by all FSK readers (HID Prox, ioProx, Pyramid - #453 ): bit decisions are identical, the soft-float double dependency is gone, and the time-sensitive demod loop is slightly faster.

### DIFF
--- a/firmware/application/src/rfid/nfctag/lf/utils/fskdemod.c
+++ b/firmware/application/src/rfid/nfctag/lf/utils/fskdemod.c
@@ -7,7 +7,7 @@
 #define PI 3.14159265358979f
 #define GOERTZEL(FREQ, SAMPLE_RATE) (2.0 * cos((2.0 * PI * FREQ) / (SAMPLE_RATE)))
 
-float goertzel_mag(float coef, uint16_t samples[], int n) {
+static float goertzel_power(float coef, const uint16_t samples[], int n) {
     float z1 = 0;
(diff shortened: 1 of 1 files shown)


LupusE (approved):
This maybe solves some effects we can't explain right now.
Thanks for sharing!

Still works at my emulated tests, I have no real Reader to test.

TeCHiScy:
also LGTM.

PR #452, opened 2026-06-24
feat(lf): DC-block FSK demod for weak-coupling robustness (HID, ioProx)

wereii:
Found out about this while working with Claude on adding my Pyramid tag #453 - the reality is that the tag does not read that well on CU and this was the deciding step to get the data demodulated properly for that type of FSK. 

---
Claude:

The FSK demod (fsk_feed) ran the Goertzel tone detection directly on the raw ADC samples. Their large DC offset leaks into the fc/8 bin and biases every decision toward bit 0; on a weakly-coupled card the AC signal is too small to overcome it, so the whole bitstream collapses to zeros and nothing decodes. Strong signals overrode the bias, which is why this went unnoticed.

Adds an opt-in running DC baseline tracker to the demod (fsk_t.dc_block): a slow per-sample IIR estimate (~64-sample time constant) subtracted before the Goertzel. Unlike a per-window mean it doesn't distort fc/8's non-integer cycle count. Enabled for the HID Prox and ioProx modems.

Validated on hardware (HID H10301, FC 12 / CN 3456 on a T55xx clone):
  - well coupled  (std 25): stock reads;        with fix 5/5
  - weak coupled  (std 15): stock fails 4/4;     with fix 4/5 correct
  - very weak     (std 11): stock fails;         with fix reads (occasional
    misread — HID 26-bit has only parity, no CRC, so marginal bit errors can
    pass; this is an HID format limitation, not introduced here)
ioProx uses the identical demod path; enabled by analogy (no card to test).

### DIFF
--- a/firmware/application/src/rfid/nfctag/lf/protocols/hidprox.c
+++ b/firmware/application/src/rfid/nfctag/lf/protocols/hidprox.c
@@ -50,6 +50,12 @@ hidprox_codec *hidprox_codec_alloc(void) {
     hidprox_codec *d = malloc(sizeof(hidprox_codec));
     d->card = NULL;
     d->modem = fsk_alloc(FSK_BITRATE_HID);
+    if (d->modem) {
+        // Subtract a running DC baseline before the Goertzel. The raw ADC DC
+        // offset biases the fc/8 bin and, on weakly-coupled cards, pins every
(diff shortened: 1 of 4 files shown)


LupusE (approved):
HidProx is still readable with PM3 after this change.

PR #459, opened 2026-07-10
#436 - Add --nopwd option for T55xx em410x write

### LINKED ISSUE #436: [idea] No Password option for T55xx
**Is your feature request related to a problem? Please describe.**
Upon writing onto T55xx cards, a default password is place on Block 0 as **20206666**, even choosing **00000000** will lock the block 0 with the password as there's no way of having no password.

**Describe the solution you'd like**
Add in a future option such as a "no password" option or a parameter with -nopwd in order to be able to read/detect/modify the tags for use or debugging purposes.

**Describe alternatives you've considered**
A clear way of creating tags or modifying tags easier via Proxmark3 or Flipper Zero without the need for a password

**Additional context**
Maybe have the nopwd as a default option and before writing a tag to optionally add a password?

naaraxi:
Fixes #436


naaraxi:
NOTE: This PR exposes the no-password option (--nopwd) for EM410X only

NOTE: Exposing this does require a client change (included here for the Python CLI)

naaraxi:
NOTE: Adding the --nopwd option to other card types should be trivial, but I figured it was outside the scope this time. I may be wrong

itsnaeng:
Just need to confirm it only works on EM410X for this for the time being?

naaraxi:
@itsnaeng 
Correct

PR #462, opened 2026-07-18
[#460] - Option to disable BLE advertising (USB-only mode)

### LINKED ISSUE #460: Option to disable BLE advertising (USB-only mode)
Hello Chameleon Ultra team!

First of all, thank you for the amazing project!

I have a question regarding the BLE functionality. Is there currently a way to completely disable BLE advertising so that the device communicates only via USB? I couldn't find an option for this in the app or the CLI commands.

My use case: I often use the device connected via USB only and would like to turn off the Bluetooth module to avoid unnecessarily cluttering the radio spectrum.

If I simply missed this feature, I apologize – could you please point me to the right command/setting?

If this is not currently possible (and assuming it's not a hardware limitation), please consider this a Feature Request. It would be incredibly useful to have a toggle or a CLI command to turn BLE on and off as needed.

Thank you for your time and hard work!

naaraxi:
resolves #460 

default is on
turning it off silences the radio and drops any live BLE link

note: you can only change it over USB, so you can't cut off the link you're using

usage:
- hw settings bleadv --- check status
- hw settings bleadv -d --- off
- hw settings bleadv -e --- on
- hw settings store --- save after enabling/disabling BLE


GameTec-live:
Stupid question: What would be wrong with cutting the active link? As long as the client shows a warning or smthng? I dont think itd break anything? Your connections just gone then until you find something USB capable.

For example iOS can only do BLE, so you cant disable BLE on iOS?

naaraxi:
@GameTec-live 

Nothing particularly wrong with it, I just figured it's cleaner this way.
If we were to remove the USB gate, and you cut the connection while you're connected via BLE, the setting is in memory and will persist until the next reboot / wake from deep sleep (at which point it will read the flash and re-enable BLE).
Saving the setting when the only way to connect to it is via BLE would essentially render it useless, so I'd advise against implementing something like that.

GameTec-live:
Oh right, because you couldnt send the save command at this point?

naaraxi:
Correct. It's possible to add a save to it, but I really wouldn't do that :D

Issue #324, opened 2025-12-05
Problem with MiFare Classic 1k emulation (MF3.2)


ZiRead:
Hi, I have a problem with emulating the MiFare Classic 1k tag. I had a new intercom installed on my front door that works with the MiFare Classic 1k tags, and I was able to read the dump completely and find all the crypto keys. The dump is working because I recorded it on the tag and checked, it works, but in the emulation mode, the intercom will not open. This intercom only works with the Gen 3.2 and Gen 4 tags. I tried to combine the emulation parameters, but it didn't lead to anything. Is there any way I can configure my Chameleon Ultra to emulate the Gen 3.2 and Gen 4 tags?

EDX2fx:
is it also an intercom system from ufanet?)

ZiRead:
This intercom was made in my country and is apparently quite modern, as it recognizes Gen 1 and Gen 2 tags. I can't view the intercom logs, but I can say that using Chameleon Ultra, I was able to copy my tag to a Gen 3.2 card. But I can't emulate my tag.

ZiRead:
I double-checked it again and was finally convinced. I can create a working copy of my tag using the Gen 3.2 blank through the MTools BLE application. But no matter how I change the emulation configuration, I can't get my Chameleon Ultra to work similarly to the Gen 3.2 tag. I don't have enough programming knowledge to add this functionality to the firmware and software myself, so I ask developers to pay attention to this. This fix would significantly expand the list of readers that Chameleon Ultra could work with in emulation mode. I thank the developers for the work they have already done, the device turned out to be very cool.

whywilson:
In fact, Gen1, Gen2, Gen3, or Gen4 are Magic Tag types, not card types.

The Mifare Classic 1K emulation by ChameleonUltra uses the standard protocol, but it can enable Gen1a or Gen2 features, which need to be disabled during normal use. Additionally, signal issues or an excessively long frame delay time (5000 vs 1000) could cause emulation failure.

I also have a question about why Gen2 type Mifare Classic 1K cards don't work; they can't be detected.

ZiRead:
[...]

PR #470, opened 2026-08-10
Make MF1 strict Key B auth optional

### LINKED ISSUE #467: The Chameleon stopped emulating the card on firmware 2.2.
Hello. I have a similar issue with an electronic door lock. On firmware 2.2, the Chameleon cannot unlock this lock, but I can write the card data to any UID/CUID (and similar) MIFARE Classic magic card. Any of these cards unlocks the door successfully, but the Chameleon itself does not.

The same lock works correctly with the Chameleon only on firmware 2.1.

Is there anything that can be done to make the Chameleon unlock this lock on firmware 2.2?

The Chameleon is not an original device. The lock I’m trying to unlock is a battery-powered electronic door lock.

unkernet:
Fixes #467 

  PR #274 tightened the MF1 emulator's authentication state machine to reject Key B authentication on sectors whose ACS bits make Key B readable via Key A.

  This is technically correct, and matches real NXP MF1 cards. However, users have reported #467 that with firmware 2.2 the Chameleon can no longer open certain door locks that worked fine with 2.1 — even when the *same dump* written to a UID/CUID/gen2 Mifare Classic magic card unlocks the same reader without any issue.

  I am introduce a per-slot config bit `mode_strict_key_b_auth`:
  | Value | Behaviour |
  | --- | --- |
  | **0** | Permissive. Accept Key B auth regardless of ACS. Matches Chinese clones, matches pre-#274 firmware |
  | **1** | Strict. Enforce the NXP MF1S50 rule, reject Key B auth when Key B is readable via Key A |

CLI additions:
`hf mf econfig --enable_strict_key_b_auth` and `hf mf econfig --disable_strict_key_b_auth` to switch this flag.


unkernet:
@GameTec-live user have reported issue with the firmware 2.2 and I hope this fix will be merged and released as 2.2.1 soon

ZiRead:
Hello, I'm sorry for the offtop, but still. Can you tell me if this PR will help with the problem we are discussing here #324?

GameTec-live:
@unkernet Good catch, if we can get confirmed tests / reviews that this fixes it, ill add it to this release cycle as a hotfix.

Whats the default setting for slots? Permissive or restrictive?

unkernet:
Default is permissive, like in was before. A have checked it against different cards I have from the random cities transport systems (all of them it not a chinese magic cards) and 3 out of 7 allowing to auth with key B in transport configuration AC bits.

jagular-spb:
@unkernet any news on this one ?

PR #487, opened 2026-09-13
Update README.md

r4zr1:
Add new ios App companion


GameTec-live:
Please get this app published to the public appstore first instead of publishing a testflight link. Thanks!

